PR #17, opened 2023-08-31
Add langchain callback

nicolewhite:
https://app.autoblocks.ai/simulations/run?runId=nicole-20230831-091639 (in our demo org)


.github/workflows/ci-vendor.yml
```diff
@@ -0,0 +1,67 @@
+name: CI - Vendor
```
nicolewhite, line 1: added a separate workflow for testing the n most recent versions of vendor packages so that we can be notified if they ever introduce breaking changes that break our callback handler

adamnolte (approved):

PR #19, opened 2023-09-04
Add more event filter operators

nicolewhite:


adamnolte (dismissed):
Just realized we have equals and not equals but then it's equal singlular for the new ones. Not sure what's standard

nicolewhite:
hmm true, want me to make the new ones `LESS_THAN_OR_EQUALS` and `GREATER_THAN_OR_EQUALS`?

nicolewhite:
or switch `EQUALS` to `EQUAL` and `NOT_EQUALS` to `NOT_EQUAL`

nicolewhite:
ClickHouse appears to use "equals" for everything: https://clickhouse.com/docs/en/sql-reference/functions/comparison-functions

adamnolte:
I say EQUALS plural in my head, but i'm fine with either but i think we should be consistent

adamnolte (approved):

PR #26, opened 2023-10-11
Send timestamp if not provided

nicolewhite:


autoblocks/_impl/tracer.py
```diff
@@ -91,6 +93,7 @@ def send_event(
         merged_properties.update(properties or {})
 
         trace_id = trace_id or self._trace_id
+        timestamp = timestamp or datetime.now(timezone.utc).isoformat()
```
adamnolte, line 96: assuming this is equivilant to `datetime.utcnow().isoformat()` which we had in the patch file?

adamnolte (approved):

autoblocks/_impl/tracer.py
nicolewhite: actually not the same (should have used this in patch) since `datetime.utcnow().isoformat()` doesn't include timezone info (but we fixed this in our ingestion pipeline a while back to handle timestamps w/o an offset included) so effectively it is the same

```
>>> print(datetime.utcnow().isoformat(), datetime.now(timezone.utc).isoformat())
2023-10-11T01:33:08.017981 2023-10-11T01:33:08.018074+00:00
```

PR #46, opened 2023-10-31
Relax dependency requirements

nicolewhite:
Motivated by https://github.com/zanieb/poetry-relax


pyproject.toml
```diff
@@ -23,6 +22,7 @@ pytest-httpx = "^0.22.0"
 langchain = "^0.0.326"
 openai = "^0.28.1"
 freezegun = "^1.2.2"
+wrapt = "^1.15.0"
```
nicolewhite, line 25: this is only used by the openai automated patch function (`trace_openai`) so moving it to dev dependencies and doing a runtime check like we are with openai

pyproject.toml
```diff
@@ -13,8 +13,7 @@ packages = [{include = "autoblocks"}]
 
 [tool.poetry.dependencies]
 python = "^3.8.1"
-httpx = "^0.24.0"
-wrapt = "^1.15.0"
+httpx = ">=0.24.0"
```
nicolewhite, line 16: relaxing this so people can install e.g. httpx==0.25.0. can't update this ourselves until https://github.com/Colin-b/pytest_httpx/issues/122 is fixed tho

adamnolte (approved):

PR #51, opened 2023-11-01
Add top-level send_event args for span_id + parent_span_id

nicolewhite:


autoblocks/_impl/vendor/openai/patch.py
```diff
@@ -13,7 +13,7 @@
 )
 
 
-def _mask_request_kwargs(kwargs: Dict) -> Dict:
+def mask_request_kwargs(kwargs: Dict) -> Dict:
```
nicolewhite, line 16: this doesn't really need to be hidden via a leading underscore since it's already in the `_impl/` folder and we only "export" `trace_openai` here: https://github.com/autoblocksai/python-sdk/blob/main/autoblocks/vendor/openai.py

autoblocks/_impl/tracer.py
```diff
@@ -75,22 +75,23 @@ def update_properties(self, properties: Dict) -> None:
         """
         self._properties.update(properties)
 
-    def send_event(
+    def _send_event_unsafe(
```
nicolewhite, line 78: adding `_send_event_unsafe` so that we're try-catching this entire function within the exposed `send_event` function. previously we were only try-catching the POST request but we should probably try-catch the whole thing including the merging of properties etc

adamnolte (approved):

PR #85, opened 2024-01-06
Add env var enum & read keys from env

nicolewhite:
Also added a `start_span` context manager to make it easier to enter / exit spans


autoblocks/_impl/tracer.py
```diff
@@ -44,10 +46,27 @@ def __init__(
...
+            return
+
+        if AutoblocksTracer._client.headers.get("authorization") != f"Bearer {ingestion_key}":
+            raise ValueError(
+                "You must use the same configuration for all AutoblocksTracer instances in your application."
+            )
+        if AutoblocksTracer._client.timeout != httpx.Timeout(timeout.total_seconds()):
+            raise ValueError(
+                "You must use the same configuration for all AutoblocksTracer instances in your application."
+            )
```
nicolewhite, line 69: this is a little awkward but probably the easiest way forward in the SDK's current state to allow creating new instances of the tracer on every request w/o creating a new client every time. likely in future versions we will want to decouple the client and the tracer

autoblocks/_impl/tracer.py
```diff
@@ -76,6 +95,27 @@ def update_properties(self, properties: Dict) -> None:
         """
         self._properties.update(properties)
 
+    @contextmanager
+    def start_span(self):
```
nicolewhite, line 99: note this will only work properly in synchronous contexts. if a user has an async app we'll need to add an async tracer and use [contextvars](https://docs.python.org/3/library/contextvars.html) to have task-level properties

PR #88, opened 2024-01-16
EPD-507 Headless prompt SDK

nicolewhite:
Users will need to add a YAML config file like:

`.autoblocks.yml`:

```yaml
autogenerate:
  prompts:
    outfile: myapp/somedir/somefile.py

    prompts:
      - id: some-prompt-a
        major_versions:
          - 1

      - id: some-prompt-b
        major_versions:
          - 2
```

Then run `autoblocks prompts generate` to autogenerate the classes they can use to execute the prompts.

This will let the CLI know which prompts to generate code for. In the typical case, the `major_versions` array will only be a single version so that we only autogenerate code for the major version you're currently using, but made it an array in case someone wants to use multiple major versions of the same prompt at once, e.g. for A/B testing.

The two classes people will actually use are the subclasses of `AutoblocksPromptManager` and the minor version enum (see the e2e tests). For example:

```python
mgr = UsedByCiDontDeletePromptManager(
    UsedByCiDontDeleteMinorVersion.v1,
)

with mgr.exec() as ctx:
    assert ctx.params.frequency_penalty == 0
    assert ctx.params.max_tokens == 256
    assert ctx.params.model == "gpt-4"
    assert ctx.params.presence_penalty == -0.3
    assert ctx.params.temperature == 0.7
    assert ctx.params.top_p == 1

    assert (
        ctx.render.template_a(
            name="Alice",
            weather="sunny",
        )
        == "Hello, Alice! The weather is sunny today."
    )
```


autoblocks/_impl/prompts/autogenerate.py
```diff
@@ -0,0 +1,327 @@
...
+    @property
+    def snake_case_id(self) -> str:
+        return to_snake_case(self.id)
+
+
+class Prompt(FrozenModel):
+    id: str
+    major_version: str
+    has_multiple_major_versions: bool
+    params: Dict[str, Any]
```
adamnolte, line 39: does params get set to an empty dictionary if it's not defined in the ui?

nicolewhite: yeah i'm setting it to an empty dictionary when parsing the API response:

```python
        try:
            params = row["params"]["params"] or {}
        except KeyError:
            params = {}
```

could also make this `Optional[Dict[str, Any]]` but has the same effect either way in the code that's produced, which will just be an empty pydantic class:

```python
    auto = f"class {prompt.params_class_name}(FrozenModel):\n"

    if not prompt.params:
        auto += f"{indent()}pass\n"
        return auto
```

nicolewhite: there is an example in the test here as well: https://github.com/autoblocksai/python-sdk/pull/88/files#diff-2575c7a671ecad6ae75e68531604c281fc160335aa67056ff0b839eeb4908d23R361-R362

adamnolte (dismissed):

autoblocks/_impl/prompts/autogenerate.py
nicolewhite: potentially a better option here though is to make the `params` property on the execution context have a None type if there are no params, would probably lead to better typing hinting rather than an empty pydantic class. let me mess with it a bit

nicolewhite: hmm yeah in pycharm at least it doesn't seem to complain if you reference an attribute that doesn't exist on an empty class, whereas it will if you try to reference an attribute on None

Screenshot 2024-01-16 at 1 19 48 PM

adamnolte (approved):

PR #89, opened 2024-01-22
Improve type hinting for prompts with no model parameters

nicolewhite:


tests/autoblocks/test_prompts_autogenerate.py
```diff
@@ -404,13 +400,17 @@ def template_c(
 
 class PromptB2ExecutionContext(
     PromptExecutionContext[
-        PromptB2Params,
+        None,
```
nicolewhite, line 403: i thought passing None for the generic here would be sufficient but in pycharm at least it doesn't recognize that it's None. ended up needing to override `params` to return None explicitly

adamnolte (approved):

PR #103, opened 2024-02-16
Add dangerously use undeployed

nicolewhite:


e2e/run.py
```diff
@@ -204,3 +206,12 @@ def test_prompt_manager_no_model_params():
...
         assert prompt.params is None
+
+
+def test_prompt_manager_no_model_params_undeployed():
+    mgr = UsedByCiDontDeleteNoParamsUndeployedPromptManager(
+        UsedByCiDontDeleteNoParamsUndeployedMinorVersion.DANGEROUSLY_USE_UNDEPLOYED,
+    )
+
+    with mgr.exec() as prompt:
+        assert prompt.params is None
```
dmorton2297, line 217: Any reason why we don't have these tests under `tests` folder in this directory?

autoblocks/_impl/prompts/autogenerate.py
```diff
@@ -328,7 +353,18 @@ def generate_code_for_config(config: AutogeneratePromptsConfig) -> str:
     resp.raise_for_status()
     data = resp.json()
 
-    prompts = make_prompts_from_api_response(data, config)
+    # Get undeployed prompts as well
+    for prompt_id in set(p.id for p in config.prompts):
+        resp = httpx.get(
+            f"{API_ENDPOINT}/prompts/{prompt_id}/major/{UNDEPLOYED}/minor/{UNDEPLOYED}",
+            headers={"Authorization": f"Bearer {AutoblocksEnvVar.API_KEY.get()}"},
+        )
```
dmorton2297, line 361: Conceptually - what does `deploying` a prompt mean? Looking at prompt management in the UI - are we adding new options there?

Struggling with the use case of why I wouldn't just use a major version of 0

dmorton2297 (dismissed):

e2e/run.py
nicolewhite: yeah we could move them there, we'd just need to refactor them a bit to be pytest-compatible. we'd also need to update the pytest config so that `poetry run pytest` ignores the `tests/e2e/` folder by default. i can create a ticket for this

autoblocks/_impl/prompts/autogenerate.py
nicolewhite: let's do a call to discuss

dmorton2297 (approved):

PR #104, opened 2024-02-17
Add testing sdk

nicolewhite:


autoblocks/_impl/testing/run.py
```diff
@@ -0,0 +1,285 @@
...
+        "/results",
+        json=dict(
+            testExternalId=test_id,
+            testCaseHash=test_case._cached_hash,
+            testCaseBody=dataclasses.asdict(test_case),
+            testCaseOutput=output,
+        ),
+    )
+
+    await gather_with_max_concurrency(
```
adamnolte, line 200: do we need to handle an exception being returned here and send it to `/errors`?

autoblocks/_impl/testing/run.py
```diff
@@ -0,0 +1,285 @@
...
+            test_id=test_id,
+            test_case_hash=None,
+            evaluator_id=None,
+            error=err,
+        )
+        return
+
+    await client.post("/start", json=dict(testExternalId=test_id))
+
+    await gather_with_max_concurrency(
```
adamnolte, line 245: same question here on handling an exception being returned

tests/autoblocks/test_run_test_suite.py
```diff
@@ -0,0 +1,693 @@
...
+def test_no_test_cases(httpx_mock):
+    httpx_mock.add_response()
+
+    run_test_suite(
+        id="my-test-id",
+        test_cases=[],
+        evaluators=[],
+        fn=lambda _: None,
+        max_test_case_concurrency=1,
+        max_evaluator_concurrency=1,
```
adamnolte, line 47: is it easy to add a test to make sure the concurrency params are respected?

adamnolte (dismissed):
Overall looks great, a couple questions but they can also be followups

autoblocks/_impl/testing/run.py
nicolewhite: that's handled in the function it's calling: 

```python
        try:
            output = await fn(test_case)
        except Exception as err:
            await send_error(
                test_id=test_id,
                test_case_hash=test_case._cached_hash,
                evaluator_id=None,
                error=err,
            )
```

and we shouldn't need to try/catch `gather_with_max_concurrency` b/c we're setting return_exceptions=True, which makes it like `Promise.allSettled`. but now that you mention it, maybe a cleaner way to do this would be to let the called function throw and then handle the exception when it's returned from `gather_with_max_concurrency`

tests/autoblocks/test_run_test_suite.py
nicolewhite: i think what i can do is just add a test that asserts on the order of the requests made, since it should be deterministic if both concurrency settings are set to 1

autoblocks/_impl/testing/run.py
nicolewhite: hmm might punt on this, looks like the stacktrace isn't propagated and gather() waits for everything to complete, would need to figure out how to propagate the stacktrace and find something that yields as things complete

adamnolte: Ya I was mainly wondering if we needed to handle asyncio throwing an error in that function

nicolewhite: ok i can add that but won't be able to add a test i don't think since we're catching all of the errors that we can mock in a test

adamnolte (approved):

autoblocks/_impl/testing/run.py
```diff
@@ -0,0 +1,301 @@
...
+    id: str,
+    test_cases: List[BaseTestCase],
+    evaluators: List[BaseEvaluator],
+    fn: Callable,
+    # How many test cases to run concurrently
+    max_test_case_concurrency: int = 10,
+    # How many evaluators to run concurrently on the result of a test case
+    max_evaluator_concurrency: int = 5,
+):
+    init_global_vars()
```
dmorton2297, line 288: nit: I think this function name should also mention the fact that it starts the event loop

autoblocks/_impl/testing/run.py
```diff
@@ -0,0 +1,301 @@
...
+                message=str(error),
+                stacktrace=traceback.format_exc(),
+            ),
+        ),
+    )
+
+
+async def gather_with_max_concurrency(
+    max_concurrency: int,
+    *coroutines: Coroutine,
```
dmorton2297, line 86: I didn't know you can use this `*` syntax here. 

Couldn't this also be `coroutines: list[Coroutine]`?

autoblocks/_impl/testing/run.py
```diff
@@ -0,0 +1,301 @@
...
+        try:
+            output = await fn(test_case)
+        except Exception as err:
+            await send_error(
+                test_id=test_id,
+                test_case_hash=test_case._cached_hash,
+                evaluator_id=None,
+                error=err,
+            )
+    else:
```
dmorton2297, line 175: These if else statements - essentially checking if `fn` was defined using the `async def` function syntax right?

tests/autoblocks/util.py
```diff
@@ -9,3 +9,7 @@ def make_expected_body(data: Dict) -> bytes:
     https://github.com/Colin-b/pytest_httpx#matching-on-http-body-1
     """
     return json.dumps(data).encode()
+
+
+def decode_request_body(req) -> Dict:
```
dmorton2297, line 14: nit: could just using built in `dict` since you aren't specifying generic values in the return type

autoblocks/_impl/testing/run.py
nicolewhite: yes, just wanted to match the original signature. but we can do that instead

autoblocks/_impl/testing/run.py
nicolewhite: yeah exactly: https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction

adamnolte (approved):

PR #105, opened 2024-02-17
Move e2e tests into tests/

nicolewhite:


tests/e2e/test_e2e.py
```diff
@@ -20,25 +19,17 @@
 from .prompts import UsedByCiDontDeleteNoParamsUndeployedPromptManager
 from .prompts import UsedByCiDontDeletePromptManager
 
-AUTOBLOCKS_API_KEY = os.environ.get("AUTOBLOCKS_API_KEY")
-AUTOBLOCKS_INGESTION_KEY = os.environ.get("AUTOBLOCKS_INGESTION_KEY")
```
dmorton2297, line 24: Why can we omit these env vars now?

nicolewhite: these tests were written before i updated the SDKs to read from env vars if available, so you don't need to explicitly pass them in anymore

PR #115, opened 2024-02-22
EPD-612: Update AutoblocksTracer to use global event loop

dmorton2297:
## Description 
Updated `tracer.py` to use the new global event loop pattern

### DIFF
--- a/autoblocks/_impl/tracer.py
+++ b/autoblocks/_impl/tracer.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import uuid
 from contextlib import contextmanager
@@ -8,24 +9,11 @@ from datetime import timezone
 from typing import Dict
 from typing import Optional
 
-import httpx
-
+from autoblocks._impl import global_state
 from autoblocks._impl.config.constants import INGESTION_ENDPOINT
 from autoblocks._impl.util import AutoblocksEnvVar
 
 log = logging.getLogger(__name__)
-
-
-global_ingestion_client = None
-
-
-def get_or_create_ingestion_client(ingestion_key: str) -> httpx.Client:
-    global global_ingestion_client
-    if global_ingestion_client is None:
-        global_ingestion_client = httpx.Client(
-            headers={"Authorization": f"Bearer {ingestion_key}"},
-        )
-    return global_ingestion_client
 
 
 @dataclass
@@ -52,6 +40,7 @@ class AutoblocksTracer:
         # Timeout for sending events to Autoblocks
         timeout: timedelta = timedelta(seconds=5),
     ):
+        global_state.init()  # Start up event loop if not already started
         self._trace_id: Optional[str] = trace_id
         self._properties: Dict = properties or {}
 
@@ -61,8 +50,12 @@ class AutoblocksTracer:
                 f"You must provide an ingestion_key or set the {AutoblocksEnvVar.INGESTION_KEY} environment variable."
             )
 
-        self._client = get_or_create_ingestion_client(ingestion_key)
+        self._ingestion_key = ingestion_key
+        self._client = global_state.http_client()
         self._timeout_seconds = timeout.total_seconds()
+
+    def _get_client_headers(self) -> Dict:
+        return {"Authorization": f"Bearer {self._ingestion_key}"}
 
     def set_trace_id(self, trace_id: str) -> None:
         """
@@ -112,7 +105,7 @@ class AutoblocksTracer:
                 props["span_id"] = prev_span_id
             self.set_properties(props)
 
-    def _send_event_unsafe(
+    async def _send_event_unsafe(
         self,
         # Require all arguments to be specified via key=value
         *,
@@ -133,19 +126,26 @@ class AutoblocksTracer:
         trace_id = trace_id or self._trace_id
         timestamp = timestamp or datetime.now(timezone.utc).isoformat()
 
-        req = self._client.post(
-            url=INGESTION_ENDPOINT,
-            json={
-                "message": message,
-                "traceId": trace_id,
-                "timestamp": timestamp,
-                "properties": merged_properties,
-            },
-            timeout=self._timeout_seconds,
-        )
-        req.raise_for_status()
-        resp = req.json()
-        return SendEventResponse(trace_id=resp.get("traceId"))
+        event_dict = {
+            "message": message,
+            "traceId": trace_id,
+            "timestamp": timestamp,
+            "properties": merged_properties,
+        }
+        try:
+            req = await self._client.post(
+                url=INGESTION_ENDPOINT,
+                json=event_dict,
+                headers=self._get_client_headers(),
+                timeout=self._timeout_seconds,
+            )
+            resp = req.json()
+            return SendEventResponse(trace_id=resp.get("traceId"))
+        except Exception as err:
+            log.error(f"Failed to send event to Autoblocks: {err}", exc_info=True)
+            if AutoblocksEnvVar.TRACER_THROW_ON_ERROR.get() == "1":
+                raise err
+            return SendEventResponse(trace_id=None)
 
     def send_event(
         self,
@@ -164,18 +164,15 @@ class AutoblocksTracer:
         Always returns a SendEventResponse dataclass as the response. If sending
         the event failed, the trace_id will be None.
         """
-        try:
-            return self._send_event_unsafe(
+        future = asyncio.run_coroutine_threadsafe(
+            self._send_event_unsafe(
                 message=message,
                 trace_id=trace_id,
                 span_id=span_id,
                 parent_span_id=parent_span_id,
                 timestamp=timestamp,
                 properties=properties,
-            )
-        except Exception as err:
-            if AutoblocksEnvVar.TRACER_THROW_ON_ERROR.get() == "1":
-                raise err
-
-            log.error(f"Failed to send event to Autoblocks: {err}", exc_info=True)
-            return SendEventResponse(trace_id=None)
+            ),
+            global_state.event_loop(),
+        )
+        return future.result()
--- a/autoblocks/_impl/tracer.py
+++ b/autoblocks/_impl/tracer.py
@@ -51,7 +51,6 @@ class AutoblocksTracer:
             )
 
         self._client_headers = {"Authorization": f"Bearer {ingestion_key}"}
-        self._client = global_state.http_client()
         self._timeout_seconds = timeout.total_seconds()
 
     def set_trace_id(self, trace_id: str) -> None:
@@ -123,19 +122,18 @@ class AutoblocksTracer:
         trace_id = trace_id or self._trace_id
         timestamp = timestamp or datetime.now(timezone.utc).isoformat()
 
-        event_dict = {
-            "message": message,
-            "traceId": trace_id,
-            "timestamp": timestamp,
-            "properties": merged_properties,
-        }
-
-        req = await self._client.post(
+        req = await global_state.http_client().post(
             url=INGESTION_ENDPOINT,
-            json=event_dict,
+            json={
+                "message": message,
+                "traceId": trace_id,
+                "timestamp": timestamp,
+                "properties": merged_properties,
+            },
             headers=self._client_headers,
             timeout=self._timeout_seconds,
         )
+        req.raise_for_status()
         resp = req.json()
         return SendEventResponse(trace_id=resp.get("traceId"))
 
--- a/autoblocks/_impl/tracer.py
+++ b/autoblocks/_impl/tracer.py
@@ -50,12 +50,9 @@ class AutoblocksTracer:
                 f"You must provide an ingestion_key or set the {AutoblocksEnvVar.INGESTION_KEY} environment variable."
             )
 
-        self._ingestion_key = ingestion_key
+        self._client_headers = {"Authorization": f"Bearer {ingestion_key}"}
         self._client = global_state.http_client()
         self._timeout_seconds = timeout.total_seconds()
-
-    def _get_client_headers(self) -> Dict:
-        return {"Authorization": f"Bearer {self._ingestion_key}"}
 
     def set_trace_id(self, trace_id: str) -> None:
         """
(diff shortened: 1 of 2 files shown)


autoblocks/_impl/testing/models.py
```diff
@@ -38,3 +38,6 @@ def id(self) -> str:
     @abc.abstractmethod
     def evaluate(self, test_case: BaseTestCase, output: Any) -> Evaluation:
         pass
+
+    def evaluateOnline(self, event: Any) -> Evaluation:
```
nicolewhite, line 42: python should be snake_case

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+
+        self._client = get_or_create_ingestion_client(ingestion_key)
+        self._timeout_seconds = timeout.total_seconds()
+        cli_server_address = AutoblocksEnvVar.CLI_SERVER_ADDRESS.get()
+        if not cli_server_address:
+            raise RuntimeError(
+                "Autoblocks tests must be run within the context of the testing CLI.\n"
+                "Make sure you are running your test command with:\n"
+                "$ npx autoblocks testing exec -- <your test command>"
+            )
```
nicolewhite, line 104: can drop this, only relevant for testing sdk

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+        self._loop = asyncio.new_event_loop()
```
nicolewhite, line 105: you'll wanna make this a global so it's only initialized once

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+        self._started = True
+
+    async def _evaluate_event(
+        self,
+        event: Any,
+        evaluator: BaseEvaluator,
+    ):
+        evaluation: Optional[Evaluation] = None
+
+        if inspect.iscoroutinefunction(evaluator.evaluate):
```
nicolewhite, line 122: shouldn't this be `evaluator.evaluate_online`?

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+                evaluation = await evaluator.evaluateOnline(event=event)
+            except Exception:
+                log.error("Do something here")
+        else:
+            try:
+                ctx = contextvars.copy_context()
+                evaluation = await loop.run_in_executor(
+                    None,
+                    ctx.run,
+                    evaluator.evaluate,
```
nicolewhite, line 133: `evaluator.evaluate_online`?

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Globals
+client: httpx.AsyncClient = None
+loop: asyncio.AbstractEventLoop = None
+started: bool = False
+
+
+# Context
+current_test_id_var = contextvars.ContextVar("current_test_id")
+current_test_case_var = contextvars.ContextVar("current_test_case")
```
nicolewhite, line 33: can drop these

autoblocks/_impl/testing/models.py
dmorton2297: 😨 lol typescript brain

autoblocks/_impl/tracerv2/run.py
dmorton2297: Copypasta 👍

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+    async def async_send_event(
+        self,
+        message: str,
+        trace_id: Optional[str] = None,
+        span_id: Optional[str] = None,
+        parent_span_id: Optional[str] = None,
+        timestamp: Optional[str] = None,
+        properties: Optional[Dict] = None,
+        evaluators: List[BaseEvaluator] = [],
+        max_evaluator_concurrency: int = 5,
```
nicolewhite, line 220: we should also make this configurable via env var

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+                span_id=span_id,
+                parent_span_id=parent_span_id,
+                timestamp=timestamp,
+                properties=properties,
+                evaluators=evaluators,
+                max_evaluator_concurrency=5,
+            ),
+            loop,
+        )
+        future.result()
```
nicolewhite, line 263: i think we probably don't want to wait here, we should just fire and forget

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+from autoblocks._impl.config.constants import INGESTION_ENDPOINT
+from autoblocks._impl.testing.models import BaseEvaluator
+from autoblocks._impl.testing.models import Evaluation
+from autoblocks._impl.tracerv2.models import SendEventResponse
+from autoblocks._impl.util import AutoblocksEnvVar
+from autoblocks._impl.util import make_replay_headers
+
+log = logging.getLogger(__name__)
+
+# Globals
```
nicolewhite, line 25: we should add a new file in like `autoblocks/global_state.py` that manages these across the entire package, e.g. for tracer, prompt manager, testing suites, etc. i think they can all share the same event loop background thread and client

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+            try:
+                ctx = contextvars.copy_context()
+                evaluation = await loop.run_in_executor(
+                    None,
+                    ctx.run,
+                    evaluator.evaluate,
+                    event,
+                )
+            except Exception:
+                log.error("#TODO do something here")
```
nicolewhite, line 137: we probably just wanna log the exception and move on, we aren't going to want to re-raise

autoblocks/_impl/testing/models.py
nicolewhite, line 42: not sure we want this on the same `BaseEvaluator` because we'll want to support all three of these:

* a user is only using offline evaluators
* a user is only using online evaluators
* a user is using both online and offline evaluators

currently this won't work for a user only using online evaluators because python will throw when they haven't implemented `evaluate`

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+current_test_id_var = contextvars.ContextVar("current_test_id")
+current_test_case_var = contextvars.ContextVar("current_test_case")
+
+
+def get_or_create_ingestion_client(ingestion_key: str) -> httpx.Client:
+    global global_ingestion_client
+    if global_ingestion_client is None:
+        global_ingestion_client = httpx.Client(
+            headers={"Authorization": f"Bearer {ingestion_key}"},
+        )
```
nicolewhite, line 41: and then for the global client in `global_state.py` we will just do: `client = httpx.AsyncClient()` and we'll set things like headers etc when making the request

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+                "properties": merged_properties,
+            },
+            headers=replay_headers,
+            timeout=self._timeout_seconds,
+        )
+        req.raise_for_status()
+        resp = req.json()
+        return SendEventResponse(trace_id=resp.get("traceId"))
+
+    async def async_send_event(
```
nicolewhite, line 211: i don't think we need this level right? can just rename `_send_event_unsafe` to `_async_send_event` and call it from the main `send_event`

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+        span_id: Optional[str] = None,
+        parent_span_id: Optional[str] = None,
+        timestamp: Optional[str] = None,
+        properties: Optional[Dict] = None,
+        evaluators: List[BaseEvaluator] = [],
+        max_evaluator_concurrency: int = 5,
+    ) -> SendEventResponse:
+        # Do some assertions
+        await gather_with_max_concurrency(
+            max_evaluator_concurrency,
```
nicolewhite, line 224: i think this only makes sense when gathering the results from each evaluator. here we're just passing in one coroutine

autoblocks/_impl/tracerv2/run.py
```diff
@@ -0,0 +1,263 @@
...
+        self,
+        message: str,
+        # Require all arguments after message to be specified via key=value
+        *,
+        trace_id: Optional[str] = None,
+        span_id: Optional[str] = None,
+        parent_span_id: Optional[str] = None,
+        timestamp: Optional[str] = None,
+        properties: Optional[Dict] = None,
+        evaluators: List[BaseEvaluator],
```
nicolewhite, line 248: this shouldn't be a required argument

autoblocks/_impl/tracerv2/run.py
nicolewhite: we'll need to make sure to attempt to wait for any outstanding futures on program exit though

dmorton2297:
@nicolewhite re-worked this PR to specifically be just changing `tracer` to use the new event loop pattern. I will work on the evaluators piece once this is merged

autoblocks/_impl/tracer.py
```diff
@@ -61,9 +50,13 @@ def __init__(
                 f"You must provide an ingestion_key or set the {AutoblocksEnvVar.INGESTION_KEY} environment variable."
             )
 
-        self._client = get_or_create_ingestion_client(ingestion_key)
+        self._ingestion_key = ingestion_key
+        self._client = global_state.http_client()
         self._timeout_seconds = timeout.total_seconds()
 
+    def _get_client_headers(self) -> Dict:
```
nicolewhite, line 57: i'd just set this once within `__init__`:

```
self._client_headers = ...
```

autoblocks/_impl/tracer.py
```diff
@@ -112,7 +105,7 @@ def start_span(self):
                 props["span_id"] = prev_span_id
             self.set_properties(props)
 
-    def _send_event_unsafe(
+    async def _send_event_unsafe(
```
nicolewhite, line 105: this was called `_send_event_unsafe` because it didn't do any error handling, and all the error handling was a level above. this allowed us to try-except the entire thing, including stuff happening before sending the request (merging properties etc)

can you update to maintain the same pattern?

autoblocks/_impl/tracer.py
```diff
@@ -133,19 +123,26 @@ def _send_event_unsafe(
...
+                headers=self._client_headers,
+                timeout=self._timeout_seconds,
+            )
+            resp = req.json()
+            return SendEventResponse(trace_id=resp.get("traceId"))
+        except Exception as err:
+            log.error(f"Failed to send event to Autoblocks: {err}", exc_info=True)
+            if AutoblocksEnvVar.TRACER_THROW_ON_ERROR.get() == "1":
+                raise err
+            return SendEventResponse(trace_id=None)
```
nicolewhite, line 145: let's just move this back to `send_event` per https://github.com/autoblocksai/python-sdk/pull/115/files#r1499906670

dmorton2297: Updated

autoblocks/_impl/tracer.py
```diff
@@ -61,7 +50,8 @@ def __init__(
                 f"You must provide an ingestion_key or set the {AutoblocksEnvVar.INGESTION_KEY} environment variable."
             )
 
-        self._client = get_or_create_ingestion_client(ingestion_key)
+        self._client_headers = {"Authorization": f"Bearer {ingestion_key}"}
+        self._client = global_state.http_client()
```
nicolewhite, line 54: we could probably drop this and just call `global_state.http_client().post` directly in send_event? wdyt?

autoblocks/_impl/tracer.py
```diff
@@ -133,17 +123,19 @@ def _send_event_unsafe(
...
-                "message": message,
-                "traceId": trace_id,
-                "timestamp": timestamp,
-                "properties": merged_properties,
-            },
+            json=event_dict,
+            headers=self._client_headers,
             timeout=self._timeout_seconds,
         )
-        req.raise_for_status()
```
nicolewhite, line 146: this still needs to raise

autoblocks/_impl/tracer.py
dmorton2297: I like that - no reason to obfuscate that its the global client

autoblocks/_impl/tracer.py
```diff
@@ -133,17 +123,19 @@ def _send_event_unsafe(
...
+        }
+
+        req = await self._client.post(
             url=INGESTION_ENDPOINT,
-            json={
-                "message": message,
-                "traceId": trace_id,
-                "timestamp": timestamp,
-                "properties": merged_properties,
-            },
```
nicolewhite, line 143: we can probably switch back to inlining the json? any reason to pull it out into `event_dict`?

autoblocks/_impl/tracer.py
```diff
@@ -165,17 +157,20 @@ def send_event(
...
+                    message=message,
+                    trace_id=trace_id,
+                    span_id=span_id,
+                    parent_span_id=parent_span_id,
+                    timestamp=timestamp,
+                    properties=properties,
+                ),
+                global_state.event_loop(),
             )
+            return future.result()
```
nicolewhite, line 171: as a follow up we should think about how we can make this fire and forget

tests/autoblocks/test_tracer.py
```diff
@@ -451,17 +448,3 @@ def test_tracer_start_span(*args, **kwargs):
...
-    {
-        "AUTOBLOCKS_INGESTION_KEY": "key",
-    },
-)
-def test_tracer_single_client():
-    tracer1 = AutoblocksTracer()
-    assert tracer1._client is not None
-    tracer2 = AutoblocksTracer()
-    assert tracer2._client is tracer1._client
-    assert id(tracer2._client) == id(tracer1._client)
```
dmorton2297, line 467: Removing this test as tracer now uses global client

nicolewhite (approved):
LGTM!

PR #117, opened 2024-02-22
EPD-612: Created global_state.py - created reusable event loop logic

dmorton2297:
Moved event loop logic to be in a `global_state` file so that it can be more easily reused

### DIFF
--- a/autoblocks/_impl/global_state.py
+++ b/autoblocks/_impl/global_state.py
@@ -20,7 +20,7 @@ def init() -> None:
     _loop = asyncio.new_event_loop()
 
     background_thread = threading.Thread(
-        target=run_event_loop,
+        target=_run_event_loop,
         args=(_loop,),
         daemon=True,
     )
@@ -29,9 +29,9 @@ def init() -> None:
     _started = True
 
 
-def run_event_loop(event_loop: asyncio.AbstractEventLoop) -> None:
-    asyncio.set_event_loop(event_loop)
-    event_loop.run_forever()
+def _run_event_loop(_event_loop: asyncio.AbstractEventLoop) -> None:
+    asyncio.set_event_loop(_event_loop)
+    _event_loop.run_forever()
 
 
 def event_loop() -> asyncio.AbstractEventLoop:
--- a/autoblocks/_impl/global_state.py
+++ b/autoblocks/_impl/global_state.py
@@ -1,0 +1,70 @@
+import asyncio
+import threading
+from typing import Coroutine
+from typing import List
+from typing import Optional
+
+import httpx
+
+_client: Optional[httpx.AsyncClient] = None
+_loop: Optional[asyncio.AbstractEventLoop] = None
+_started: bool = False
+
+
+def init(client: httpx.AsyncClient) -> None:
+    global _client, _loop, _started
+
+    if _started:
+        return
+
+    _client = client
+
+    _loop = asyncio.new_event_loop()
+
+    background_thread = threading.Thread(
+        target=run_event_loop,
+        args=(_loop,),
+        daemon=True,
+    )
+    background_thread.start()
+
+    _started = True
+
+
+def run_event_loop(_loop: asyncio.AbstractEventLoop) -> None:
+    asyncio.set_event_loop(_loop)
+    _loop.run_forever()
+
+
+def event_loop() -> asyncio.AbstractEventLoop:
+    if not _loop:
+        raise Exception("Event loop not initialized")
+    return _loop
+
+
+async def gather_with_max_concurrency(
+    max_concurrency: int,
+    coroutines: List[Coroutine],
+) -> None:
+    """
+    Borrowed from https://stackoverflow.com/a/61478547
+    """
+    semaphore = asyncio.Semaphore(max_concurrency)
+
+    async def sem_coro(coro: Coroutine):
+        async with semaphore:
+            return await coro
+
+    # return_exceptions=True causes exceptions to be returned as values instead
+    # of propagating them to the caller. this is similar in behavior to Promise.allSettled
+    await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
+
+
+def http_client() -> httpx.AsyncClient:
+    if not _client:
+        raise Exception("HTTP client not initialized")
+    return _client
+
+
+def started() -> bool:
+    return _started
--- a/autoblocks/_impl/global_state.py
+++ b/autoblocks/_impl/global_state.py
@@ -1,7 +1,5 @@
 import asyncio
 import threading
-from typing import Coroutine
-from typing import List
 from typing import Optional
 
 import httpx
@@ -11,13 +9,13 @@ _loop: Optional[asyncio.AbstractEventLoop] = None
 _started: bool = False
 
 
-def init(client: httpx.AsyncClient) -> None:
+def init() -> None:
     global _client, _loop, _started
 
     if _started:
         return
 
-    _client = client
+    _client = httpx.AsyncClient()
 
     _loop = asyncio.new_event_loop()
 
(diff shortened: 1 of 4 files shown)


autoblocks/_impl/global_state.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Optional
+
+import httpx
+
+_client: Optional[httpx.AsyncClient] = None
+_loop: Optional[asyncio.AbstractEventLoop] = None
+_started: bool = False
+
+
+def init(client: httpx.AsyncClient) -> None:
```
nicolewhite, line 14: this shouldn't take any arguments, it should just create a client. otherwise it's not global

autoblocks/_impl/global_state.py
```diff
@@ -0,0 +1,70 @@
...
+    await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
+
+
+def http_client() -> httpx.AsyncClient:
+    if not _client:
+        raise Exception("HTTP client not initialized")
+    return _client
+
+
+def started() -> bool:
```
nicolewhite, line 69: this doesn't need to be exposed

autoblocks/_impl/global_state.py
```diff
@@ -0,0 +1,70 @@
...
+    _loop.run_forever()
+
+
+def event_loop() -> asyncio.AbstractEventLoop:
+    if not _loop:
+        raise Exception("Event loop not initialized")
+    return _loop
+
+
+async def gather_with_max_concurrency(
```
nicolewhite, line 45: this doesn't belong here, can just move it to the generic util file

autoblocks/_impl/global_state.py
```diff
@@ -0,0 +1,70 @@
...
+        target=run_event_loop,
+        args=(_loop,),
+        daemon=True,
+    )
+    background_thread.start()
+
+    _started = True
+
+
+def run_event_loop(_loop: asyncio.AbstractEventLoop) -> None:
```
nicolewhite, line 34: i'd rename the `_loop` arg to something else so that it doesn't have the same name as the global

autoblocks/_impl/testing/run.py
```diff
@@ -2,63 +2,31 @@
...
 from typing import Any
 from typing import Callable
-from typing import Coroutine
 from typing import List
 from typing import Optional
 
 import httpx
 import orjson
 
+import autoblocks._impl.global_state
```
nicolewhite, line 14:
```suggestion
from autoblocks._impl import global_state
```

autoblocks/_impl/testing/run.py
```diff
@@ -2,63 +2,31 @@
...
-    )
-    background_thread.start()
-
-    started = True
+# Event loop helpers
+init_event_loop = autoblocks._impl.global_state.init
+event_loop = autoblocks._impl.global_state.event_loop
+event_loop_started = autoblocks._impl.global_state.started
+http_client = autoblocks._impl.global_state.http_client
+gather_with_max_concurrency = autoblocks._impl.global_state.gather_with_max_concurrency
```
nicolewhite, line 29: you can drop all of this, just import global state and use it directly

autoblocks/_impl/testing/run.py
```diff
@@ -300,7 +250,16 @@ def run_test_suite(
...
-    init_global_vars_and_event_loop()
+    cli_server_address = AutoblocksEnvVar.CLI_SERVER_ADDRESS.get()
+    if not cli_server_address:
+        raise RuntimeError(
+            "Autoblocks tests must be run within the context of the testing CLI.\n"
+            "Make sure you are running your test command with:\n"
+            "$ npx autoblocks testing exec -- <your test command>"
+        )
+
+    if not event_loop_started():
```
nicolewhite, line 261: we should just call `global_state.init()`. init takes care of checking whether or not it's already started

autoblocks/_impl/testing/run.py
```diff
@@ -300,7 +250,16 @@ def run_test_suite(
...
+        init_event_loop(client=httpx.AsyncClient(base_url=cli_server_address))
```
nicolewhite, line 262: you're going to need to drop this arg and refactor this to pass the base_url on every call

nicolewhite: alternatively you could ignore the global client for the testing sdk and just use your own global client like it was setup previously. probably not super important for the testing sdk to use the package-wide global client

autoblocks/_impl/testing/run.py
```diff
@@ -300,8 +243,16 @@ def run_test_suite(
...
+    cli_server_address = AutoblocksEnvVar.CLI_SERVER_ADDRESS.get()
+    if not cli_server_address:
+        raise RuntimeError(
+            "Autoblocks tests must be run within the context of the testing CLI.\n"
+            "Make sure you are running your test command with:\n"
+            "$ npx autoblocks testing exec -- <your test command>"
+        )
 
+    global_state.init()
+    global_state.http_client().base_url = cli_server_address
```
nicolewhite, line 255: this is a package-wide global, you can't modify it like this. it would affect users consumers of the client

autoblocks/_impl/testing/run.py
dmorton2297: I ended up reworking it so the run_test_suit looks like

```python
def run_test_suite(
    ...stuff...
):
    cli_server_address = AutoblocksEnvVar.CLI_SERVER_ADDRESS.get()
   ...stuff

    global_state.init()
    global_state.http_client().base_url = cli_server_address
    future = asyncio.run_coroutine_threadsafe(
        ...stuff..
        global_state.event_loop(),
    )
    future.result()
```

autoblocks/_impl/testing/run.py
dmorton2297: Ah yeah your right. Maybe client doesn't need to exist here at all - its pretty light and conceptually doesn't relate much to the event_loop

nicolewhite: we still want a global http client, especially for the pieces used in production like the tracer and prompt sdk

https://www.python-httpx.org/advanced/clients/

also it doesn't need to relate to the event loop. this file is just to maintain package-wide globals

nicolewhite: i am saying for testing in particular you can either refactor so that you can use the package-wide http client or use your own client (still a global, but isolated to testing) like it was set up previously

dmorton2297: Ended up going to the latter option - still have client in `global_state` but added a different client to the testing

autoblocks/_impl/global_state.py
```diff
@@ -0,0 +1,46 @@
...
+        target=run_event_loop,
+        args=(_loop,),
+        daemon=True,
+    )
+    background_thread.start()
+
+    _started = True
+
+
+def run_event_loop(_event_loop: asyncio.AbstractEventLoop) -> None:
```
nicolewhite, line 32: we should probably add an underscore here since it's not meant to be used outside of this module

autoblocks/_impl/testing/run.py
```diff
@@ -2,63 +2,41 @@
...
 
+from autoblocks._impl import global_state
 from autoblocks._impl.testing.models import BaseEvaluator
 from autoblocks._impl.testing.models import BaseTestCase
 from autoblocks._impl.testing.models import Evaluation
 from autoblocks._impl.util import AutoblocksEnvVar
+from autoblocks._impl.util import gather_with_max_concurrency
 
-# Globals
+# httpx client
```
nicolewhite, line 21: probably still worth leaving the word "global" in this comment, e.g. "Global httpx client"

autoblocks/_impl/testing/run.py
```diff
@@ -300,8 +260,9 @@ def run_test_suite(
     # How many evaluators to run concurrently on the result of a test case
     max_evaluator_concurrency: int = 5,
 ):
-    init_global_vars_and_event_loop()
 
+    init_client()
```
nicolewhite, line 264: can you drop the extra space at line 263

dmorton2297: Dropped

autoblocks/_impl/testing/run.py
dmorton2297: Added Global

autoblocks/_impl/global_state.py
dmorton2297: Updated

nicolewhite (approved):
LGTM

PR #124, opened 2024-02-26
EPD-612: Evaluations with AutoblocksTracer

dmorton2297:
## Description
Updated tracer to accept `evaluators` as top level property, then execute them.

**Note**: I am planning on creating a folder for `tracer` under `_impl` directory. I have held off on doing this to make the diff easier to look over during the initial reviews, I will add this change once I get the majority of feedback on the changes in `tracer.py`

Graceful exit demonstration: https://drive.google.com/file/d/1ITe8iFdXY8qBAdUb3yBOPvyfNU9iJwkv/view?usp=sharing


autoblocks/_impl/testing/models.py
```diff
@@ -1,10 +1,28 @@
 import abc
 import dataclasses
 import functools
+import uuid
 from typing import Any
 from typing import Optional
 
 
+@dataclasses.dataclass()
```
dmorton2297, line 9: Note: This is temporarily here - I am going to create a tracer folder under _impl that contains its own models, but have held off to make the diff easier to review

autoblocks/_impl/util.py
```diff
@@ -50,4 +50,4 @@ async def sem_coro(coro: Coroutine):
 
     # return_exceptions=True causes exceptions to be returned as values instead
     # of propagating them to the caller. this is similar in behavior to Promise.allSettled
-    await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
+    return await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
```
dmorton2297, line 53: Add return so I can gather evaluator events as they finish

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +31,26 @@ class Threshold:
...
+        return dict(
+            evaluatorExternalId=event_evaluation.evaluator_external_id,
+            id=str(event_evaluation.id),
+            score=event_evaluation.score,
+            metadata=dict(event_evaluation.metadata) if event_evaluation.metadata else None,
+            threshold=dict(event_evaluation.threshold) if event_evaluation.threshold else None,
+        )
+
+
+# TODO: Rename TestEvaluation?
```
dmorton2297, line 53: @nicolewhite

nicolewhite: why wouldn't these be the same thing?

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +31,26 @@ class Threshold:
     gte: Optional[float] = None
 
 
+@dataclasses.dataclass()
+class EventEvaluation:
```
nicolewhite, line 35: if this is the schema returned by `evaluate_event`, this is what the end user will interact with. we shouldn't ask them to attach the `id` and `evaluator_external_id` themselves, we should that ourselves behind the scenes

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
                 props["span_id"] = prev_span_id
             self.set_properties(props)
 
+    async def evaluate_event(self, event: TracerEvent, evaluator: BaseEventEvaluator) -> None:
```
nicolewhite, line 112: this signature doesn't look right, you have it returning `None` but it returns the evaluation?

nicolewhite, line 112: shouldn't this have an underscore in front? it's not meant to be used by an end user right?

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+def test_tracer_prod_evaluations(httpx_mock):
+    test_evaluation_id = uuid.uuid4()
+
+    class MyEvaluator(BaseEventEvaluator):
+        id = "my-evaluator"
+
+        def evaluate_event(self, event: TracerEvent) -> EventEvaluation:
+            return EventEvaluation(
+                evaluator_external_id=self.id,
+                id=test_evaluation_id,
```
nicolewhite, line 465: end users shouldn't need to implement this glue for us, we should do it ourselves. this should return the same type of `Evaluation` schema as offline evaluators

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+                score=0.9,
+                threshold={"gte": 0.5},
```
nicolewhite, line 467: shouldn't this be a dataclass?

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+                message="my-message",
+                traceId="my-trace-id",
+                timestamp=timestamp,
+                properties={},
+            )
+        ),
+    )
+    tracer = AutoblocksTracer("mock-ingestion-key")
+    resp = tracer.send_event("my-message", **mock_input)
+    assert resp.trace_id == "my-trace-id"
```
nicolewhite, line 540: we need to make send_event fire and forget so that it doesn't block on evaluators

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+
+def test_tracer_prod_evaluations(httpx_mock):
+    test_evaluation_id = uuid.uuid4()
+
+    class MyEvaluator(BaseEventEvaluator):
+        id = "my-evaluator"
+
+        def evaluate_event(self, event: TracerEvent) -> EventEvaluation:
+            return EventEvaluation(
+                evaluator_external_id=self.id,
```
nicolewhite, line 464: this is also the same thing i was getting at in another PR, where we shouldn't leak internal implementation details to schemas used by end users

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +31,26 @@ class Threshold:
...
+    evaluator_external_id: str
+    score: float
+    id: Optional[str] = dataclasses.field(default_factory=lambda: str(uuid.uuid4()))
+    metadata: Optional[dict] = None
+    threshold: Optional[Threshold] = None
+
+    @classmethod
+    def to_json(cls, event_evaluation):
```
nicolewhite, line 43: add types

autoblocks/_impl/testing/models.py
```diff
@@ -1,10 +1,28 @@
...
 
+@dataclasses.dataclass()
+class TracerEvent:
+    message: str
+    trace_id: str
+    timestamp: str
+    properties: dict
+
+    @classmethod
+    def to_json(cls, event):
```
nicolewhite, line 17: add types

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
...
+        """
+        Evaluates an event using a provided evaluator.
+        """
+        if inspect.iscoroutinefunction(evaluator.evaluate_event):
+            try:
+                evaluation = await evaluator.evaluate_event(event=event)
+            except Exception as err:
+                log.error("Event evaluation through an exception", err)
```
nicolewhite, line 120:
```suggestion
                log.error("Event evaluation threw an exception", err)
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
...
+            try:
+                ctx = contextvars.copy_context()
+                evaluation = await global_state.event_loop().run_in_executor(
+                    None,
+                    ctx.run,
+                    evaluator.evaluate_event,
+                    event,
+                )
+            except Exception as err:
+                log.error("Event evaluation through an exception", err)
```
nicolewhite, line 131:
```suggestion
                log.error("Event evaluation threw an exception", err)
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
...
+
+        if evaluation is None:
+            return
+        return evaluation
```
nicolewhite, line 135:
```suggestion
        return evaluation
```

dmorton2297: 🤦

autoblocks/_impl/tracer.py
dmorton2297: 🤦

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
...
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            if evaluations and len(evaluations) > 0:
+                evaluations_json = [
+                    EventEvaluation.to_json(evaluation)
+                    for evaluation in filter(lambda x: isinstance(x, EventEvaluation), evaluations)
```
nicolewhite, line 157:
```suggestion
                    for evaluation in evaluations if isinstance(evaluation, EventEvaluation)
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,60 @@ def start_span(self):
...
+                max_evaluator_concurrency,
+                [
+                    self.evaluate_event(
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            if evaluations and len(evaluations) > 0:
```
nicolewhite, line 154: you can drop this line i think

autoblocks/_impl/tracer.py
```diff
@@ -125,14 +189,19 @@ async def _send_event_unsafe(
...
 
+        event = TracerEvent(
+            message=message,
+            trace_id=trace_id,
+            timestamp=timestamp,
+            properties=merged_properties,
+        )
+
+        transformed_event_json = await self._run_and_build_evals_properties(
+            evaluators, event, max_evaluator_concurrency
```
nicolewhite, line 198: nit: i like to use kwargs typically rather than positional args:

```python
transformed_event_json = await self._run_and_build_evals_properties(
  evaluators=evaluators,
  event=event,
  max_evaluator_concurrency=max_evaluator_concurrency,
)
```

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +31,26 @@ class Threshold:
...
+    threshold: Optional[Threshold] = None
+
+    @classmethod
+    def to_json(cls, event_evaluation):
+        return dict(
+            evaluatorExternalId=event_evaluation.evaluator_external_id,
+            id=str(event_evaluation.id),
+            score=event_evaluation.score,
+            metadata=dict(event_evaluation.metadata) if event_evaluation.metadata else None,
+            threshold=dict(event_evaluation.threshold) if event_evaluation.threshold else None,
```
nicolewhite, line 49: `dataclasses.asdict(event_evaluation.threshold)`

autoblocks/_impl/testing/models.py
dmorton2297: After reviewing the other related comments - 

I'll update with just extending the `Evaluation` class then working on obfuscating the `evaluatorExternalId` stuff

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +31,26 @@ class Threshold:
...
+    metadata: Optional[dict] = None
+    threshold: Optional[Threshold] = None
+
+    @classmethod
+    def to_json(cls, event_evaluation):
+        return dict(
+            evaluatorExternalId=event_evaluation.evaluator_external_id,
+            id=str(event_evaluation.id),
+            score=event_evaluation.score,
+            metadata=dict(event_evaluation.metadata) if event_evaluation.metadata else None,
```
nicolewhite, line 48: i think you can just pass `event_evaluation.metadata` directly? isn't metadata just `Optional[dict]`?

autoblocks/_impl/testing/models.py
nicolewhite: why does it need to be extended? shouldn't it be the same thing?

autoblocks/_impl/util.py
nicolewhite: can you update the types on the signature

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+                id=test_evaluation_id,
+                score=0.9,
+                threshold={"gte": 0.5},
+            )
+
+    mock_input = {
```
nicolewhite, line 470: can you just pass these as args directly down at send_event, hard to follow when all the args are up here

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +451,90 @@ def test_tracer_start_span(*args, **kwargs):
...
+
+
+def test_tracer_failing_evaluation(httpx_mock):
+    class MyEvaluator(BaseEventEvaluator):
+        id = "my-evaluator"
+
+        def evaluate_event(self, event: TracerEvent) -> EventEvaluation:
+            raise Exception("Something terrible went wrong")
+
+    mock_input = {
```
nicolewhite, line 515: pass these directly to send_event below

autoblocks/_impl/testing/models.py
dmorton2297: I just need to be able to add an optional `metadata` dict property here

autoblocks/_impl/tracer.py
```diff
@@ -150,6 +217,8 @@ def send_event(
         parent_span_id: Optional[str] = None,
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
+        evaluators: List[BaseEventEvaluator] = [],
```
nicolewhite, line 220: mutable default args are not recommended in python, can you default to None and then turn it into a list:

```
evaluators: Optional[List[BaseEventEvaluator]] = None,

...

evaluators=evaluators or [],
```

autoblocks/_impl/testing/models.py
nicolewhite: yeah we want that for both online and offline evaluations, so just update the current one

autoblocks/_impl/tracer.py
```diff
@@ -112,6 +172,8 @@ async def _send_event_unsafe(
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
         prompt_tracking: Optional[Dict] = None,
+        evaluators: List[BaseEventEvaluator] = [],
```
nicolewhite, line 175:
```suggestion
        evaluators: List[BaseEventEvaluator],
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,58 @@ def start_span(self):
...
+                    event,
+                )
+            except Exception as err:
+                log.error("Event evaluation threw an exception", err)
+
+        return evaluation
+
+    async def _run_and_build_evals_properties(
+        self, evaluators: BaseEventEvaluator, event: TracerEvent, max_evaluator_concurrency: int
+    ) -> List[EventEvaluation]:
```
nicolewhite, line 137: this type signature does not look right

autoblocks/_impl/testing/models.py
dmorton2297: Yup - bad wording on my part - by extend I meant update

autoblocks/_impl/util.py
dmorton2297: Added type `Future[list]` - I don't think we can get more specific than that

autoblocks/_impl/tracer.py
```diff
@@ -125,14 +187,19 @@ async def _send_event_unsafe(
         trace_id = trace_id or self._trace_id
         timestamp = timestamp or datetime.now(timezone.utc).isoformat()
 
+        event = TracerEvent(
+            message=message,
+            trace_id=trace_id,
+            timestamp=timestamp,
+            properties=merged_properties,
+        )
```
nicolewhite, line 197: i would make it so this part doesn't happen unless there are evaluators:

```python
if evaluators:
    event_to_evaluate = TracerEvent(
        message=message,
        trace_id=trace_id,
        timestamp=timestamp,
        properties=merged_properties,
    )
    evaluations_json = await self._run_evaluators(
      event=event_to_evaluate,
      evaluators=evaluators,
      max_evaluator_concurrency=max_evaluator_concurrency,
    )
    if evaluations_json:
        merged_properties["evaluations"] = evaluations_json

...
```

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +452,82 @@ def test_tracer_start_span(*args, **kwargs):
...
+        match_content=make_expected_body(
+            dict(
+                message="my-message",
+                traceId="my-trace-id",
+                timestamp=timestamp,
+                properties={
+                    "evaluations": [
+                        {
+                            "evaluatorExternalId": "my-evaluator",
+                            "id": str(test_evaluation_id),
```
nicolewhite, line 486: since this UUID is going to be generated internally and not by the end user, you can test it by mocking the uuid library https://github.com/autoblocksai/python-sdk/blob/3885382637244b262df41c33a34fba4fc39e777e/tests/autoblocks/test_tracer.py#L419-L423

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +452,82 @@ def test_tracer_start_span(*args, **kwargs):
...
     assert tracer._properties.get("parent_span_id") is None
+
+
+def test_tracer_prod_evaluations(httpx_mock):
+    test_evaluation_id = uuid.uuid4()
+
+    class MyEvaluator(BaseEventEvaluator):
+        id = "my-evaluator"
+
+        def evaluate_event(self, event: TracerEvent) -> EventEvaluation:
```
nicolewhite, line 463: can you also add a test for an evaluator with an async `evaluate_event`

autoblocks/_impl/testing/models.py
```diff
@@ -1,10 +1,29 @@
...
 
+@dataclasses.dataclass()
+class TracerEvent:
+    message: str
+    trace_id: str
+    timestamp: str
+    properties: dict
+
+    @classmethod
+    def to_json(cls, event: Dict[str, any]):
```
nicolewhite, line 18: This type is not correct, event is an instance of TracerEvent. Also this doesn't need to be a class method, it should just act on the instance (self)

autoblocks/_impl/testing/models.py
```diff
@@ -13,6 +32,26 @@ class Threshold:
...
+@dataclasses.dataclass()
+class EventEvaluation:
+    evaluator_external_id: str
+    score: float
+    id: Optional[str] = dataclasses.field(default_factory=lambda: str(uuid.uuid4()))
+    metadata: Optional[dict] = None
+    threshold: Optional[Threshold] = None
+
+    @classmethod
+    def to_json(cls, event_evaluation: Dict[str, any]):
```
nicolewhite, line 44: Same thing here, just call to_json on the instance directly

nicolewhite: The type is also wrong

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,52 @@ def start_span(self):
...
+    async def _evaluate_event(self, event: TracerEvent, evaluator: BaseEventEvaluator) -> Optional[EventEvaluation]:
+        """
+        Evaluates an event using a provided evaluator.
+        """
+        evaluation = None
+        if inspect.iscoroutinefunction(evaluator.evaluate_event):
+            try:
+                evaluation = await evaluator.evaluate_event(event=event)
+            except Exception as err:
+                log.error("Event evaluation threw an exception", err)
```
nicolewhite, line 121:
```suggestion
                log.error(f"Evaluator '{evaluator.id}' failed: {err}", exc_info=True)
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,52 @@ def start_span(self):
...
+            try:
+                ctx = contextvars.copy_context()
+                evaluation = await global_state.event_loop().run_in_executor(
+                    None,
+                    ctx.run,
+                    evaluator.evaluate_event,
+                    event,
+                )
+            except Exception as err:
+                log.error("Event evaluation threw an exception", err)
```
nicolewhite, line 132:
```suggestion
                log.error(f"Evaluator '{evaluator.id}' failed: {err}", exc_info=True)
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,52 @@ def start_span(self):
...
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            return list(filter(lambda e: isinstance(e, EventEvaluation), evaluations))
+
+        except Exception as err:
+            log.error("Unable to complete evaluating event", err)
```
nicolewhite, line 155: Screenshot 2024-02-27 at 9 10 32 AM

the first argument should be the entire message. if you want to pass `err` separately like this you can use `%s` as a placeholder

autoblocks/_impl/tracer.py
```diff
@@ -112,27 +166,49 @@ async def _send_event_unsafe(
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
         prompt_tracking: Optional[Dict] = None,
+        evaluators: List[BaseEventEvaluator] = None,
```
nicolewhite, line 169:
```suggestion
        evaluators: Optional[List[BaseEventEvaluator]] = None,
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,52 @@ def start_span(self):
...
+                max_evaluator_concurrency,
+                [
+                    self._evaluate_event(
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            return list(filter(lambda e: isinstance(e, EventEvaluation), evaluations))
```
nicolewhite, line 152:
```suggestion
            return [e for e in evaluations if isinstance(e, EventEvaluation)]
```

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +109,52 @@ def start_span(self):
...
+
+        return evaluation
+
+    async def _run_evaluators(
+        self, evaluators: BaseEventEvaluator, event: TracerEvent, max_evaluator_concurrency: int
+    ) -> List[EventEvaluation]:
+        if len(evaluators) == 0:
+            return []
+        try:
+            evaluations: List[EventEvaluation] = await gather_with_max_concurrency(
```
nicolewhite, line 142: is this type hint correct? the filter below suggests this isn't a list of EventEvaluation

autoblocks/_impl/tracer.py
```diff
@@ -112,27 +166,49 @@ async def _send_event_unsafe(
...
+                    properties=merged_properties,
+                ),
+                evaluators=evaluators,
+                max_evaluator_concurrency=max_evaluator_concurrency,
+            )
+            # We still need to check if there are any results,
+            # because evaluators can potentially fail and not resolve.
+            if evaluations:
+                # Update merged properties with computed evaluations property
+                merged_properties["evaluations"] = [EventEvaluation.to_json(x) for x in evaluations]
```
nicolewhite, line 201: once you drop the classmethod this can just be:

```
merged_properties["evaluations"] = [e.to_json() for e in evaluations]
```

autoblocks/_impl/tracer.py
```diff
@@ -150,6 +226,8 @@ def send_event(
         parent_span_id: Optional[str] = None,
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
+        evaluators: List[BaseEventEvaluator] = None,
```
nicolewhite, line 229:
```suggestion
        evaluators: Optional[List[BaseEventEvaluator]] = None,
```

autoblocks/_impl/util.py
```diff
@@ -38,7 +38,7 @@ def encode_uri_component(s: str) -> str:
 async def gather_with_max_concurrency(
     max_concurrency: int,
     coroutines: List[Coroutine],
-) -> None:
+):
```
nicolewhite, line 41: why drop this? can you update the type signature now that you're returning

autoblocks/_impl/tracer.py
```diff
@@ -1,26 +1,48 @@
...
 
-@dataclass
+@dataclasses.dataclass()
 class SendEventResponse:
     trace_id: Optional[str]
 
 
+class ClickHouseEvaluationObject(Evaluation):
+    @classmethod
+    def to_json(cls, event_evaluation: Evaluation, evaluator_external_id: str) -> Dict[str, Any]:
```
dmorton2297, line 32: I think a class method is appropriate here - I will never build an instance of this class directly. It's sole purpose is to combine this data and spit out valid json so I can use it ultimately in the events `evaluations` array

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +123,70 @@ def start_span(self):
...
+        if inspect.iscoroutinefunction(evaluator.evaluate_event):
+            try:
+                evaluation = await evaluator.evaluate_event(event=event)
+            except Exception as err:
+                return Evaluation(
+                    score=0,
+                    metadata={
+                        "error": "Evaluation raised an exception during execution",
+                        "errorDetail": f"{repr(err)}",
+                    },
```
dmorton2297, line 140: I updated the implementation so that evaluators that fail to run will still report a score of 0 with some metadata indicating it was an error. This way it will be obvious is evaluations are failing to run when inspecting traces

autoblocks/_impl/util.py
dmorton2297: Waiting till we merge change to drop python 3.8

autoblocks/_impl/tracer.py
nicolewhite: i would punt on this and continue logging errors, i think we need to come up with a better design and handle them in a first class way

autoblocks/_impl/tracer.py
nicolewhite: then i feel we don't need a class for this at all, can't we just build the json directly?

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +123,70 @@ def start_span(self):
...
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            # Join evaluation results with their evaluator ids
+            # if results is not of type Evaluation add an evaluation
+            # with score 0 and metadata with the exception message
+            return [
+                ClickHouseEvaluationObject.to_json(event_evaluation=item, evaluator_external_id=evaluators[index].id)
```
nicolewhite, line 182: can you add a test w/ multiple evaluators and ensure they're matching up as you expect

autoblocks/_impl/tracer.py
```diff
@@ -101,6 +123,70 @@ def start_span(self):
...
+
+        return evaluation
+
+    async def _run_evaluators(
+        self, evaluators: List[BaseEventEvaluator], event: TracerEvent, max_evaluator_concurrency: int
+    ) -> List[ClickHouseEvaluationObject]:
+        if len(evaluators) == 0:
+            return []
+        try:
+            evaluations: List[Evaluation] = await gather_with_max_concurrency(
```
nicolewhite, line 168: just fyi i verified `asyncio.gather` returns stuff in the same order given:

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

> If all awaitables are completed successfully, the result is an aggregate list of returned values. The order of result values corresponds to the order of awaitables in aws.

autoblocks/_impl/tracer.py
```diff
@@ -112,27 +198,49 @@ async def _send_event_unsafe(
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
         prompt_tracking: Optional[Dict] = None,
+        evaluators: Optional[List[BaseEventEvaluator]] = None,
```
nicolewhite, line 201: we can actually drop the default args (all the Nones) for everything in this function since we pass everything explicitly when calling from `send_event`

Blaso13 (approved):

tests/autoblocks/test_tracer.py
dmorton2297: Updated this and adjusted tests

dmorton2297:
@nicolewhite Ok updated so that

1. sendEvent is now fire and forget (tests updated to reflect)
2. Removed unnecessary types (like ClickhouseEvaluationObject)
3. Logging evaluator runtime exceptions and excluding the from evaluations computed property
4. General code cleanup

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+            id=str(uuid.uuid4()),
+            score=evaluation.score,
+            threshold=dataclasses.asdict(evaluation.threshold) if evaluation.threshold else None,
+            metadata=evaluation.metadata,
+            evaluatorExternalId=evaluator_external_id,
+        )
+
+    async def _evaluate_event(
+        self, event: TracerEvent, evaluator: BaseEventEvaluator
+    ) -> Optional[Union[Evaluation, None]]:
```
nicolewhite, line 124:
```suggestion
    ) -> Optional[Evaluation]:
```

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+        """
+        Evaluates an event using a provided evaluator.
+        """
+        evaluation = None
+        if inspect.iscoroutinefunction(evaluator.evaluate_event):
+            try:
+                evaluation = await evaluator.evaluate_event(event=event)
+            except Exception as err:
+                log.error("Unable to execute evaluation. Error: %s", repr(err), exc_info=True)
```
nicolewhite, line 133: do you need `repr`? i think it takes care of stringifying for you

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+            try:
+                ctx = contextvars.copy_context()
+                evaluation = await global_state.event_loop().run_in_executor(
+                    None,
+                    ctx.run,
+                    evaluator.evaluate_event,
+                    event,
+                )
+            except Exception as err:
+                log.error("Unable to execute evaluation. Error: %s", repr(err), exc_info=True)
```
nicolewhite, line 144: can we include the evaluator ID in both of these messages

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+
+        return evaluation
+
+    async def _run_evaluators(
+        self, evaluators: List[BaseEventEvaluator], event: TracerEvent, max_evaluator_concurrency: int
+    ) -> List[Dict[str, Any]]:
+        if len(evaluators) == 0:
+            return []
+        try:
+            evaluations: List[Union[Evaluation, None]] = await gather_with_max_concurrency(
```
nicolewhite, line 154:
```suggestion
            evaluations: List[Optional[Evaluation]] = await gather_with_max_concurrency(
```

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            return [
+                self._evaluation_to_json(event_evaluation=item, evaluator_external_id=evaluator.id)
+                for evaluator, item in zip(evaluators, evaluations)
+                if item is not None
```
nicolewhite, line 167: nit: rename `item` to `evaluation`

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+                    self._evaluate_event(
+                        event=event,
+                        evaluator=evaluator,
+                    )
+                    for evaluator in evaluators
+                ],
+            )
+            return [
+                self._evaluation_to_json(event_evaluation=item, evaluator_external_id=evaluator.id)
+                for evaluator, item in zip(evaluators, evaluations)
```
nicolewhite, line 166: love a good use of `zip` 😍

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
+                if item is not None
+            ]
+
+        except Exception as err:
+            log.error("Unable to complete evaluating events. Error: %s", str(err), exc_info=True)
```
nicolewhite, line 171: up above you're using `repr` instead of `str`, can we be consistent? also do we need either?

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +110,119 @@ def start_span(self):
...
             merged_properties["parent_span_id"] = parent_span_id
         if prompt_tracking:
             merged_properties["promptTracking"] = prompt_tracking
-
         trace_id = trace_id or self._trace_id
         timestamp = timestamp or datetime.now(timezone.utc).isoformat()
 
+        # If there are evaluators, run them and compute the evaluations property
+        if evaluators:
+            evaluations = await self._run_evaluators(
```
nicolewhite, line 203: i think we actually need a try-except on this as well, since we're only catching failures in the user-defined `evaluate_event`. but if there's an error in our code it will prevent the event from being sent

for example, maybe add a test where our `_evaluation_to_json` method raises an error and assert that the event is still sent

nicolewhite: in python you can use mock.patch to patch the tracer's method to raise

autoblocks/_impl/tracer.py
```diff
@@ -150,6 +240,8 @@ def send_event(
         parent_span_id: Optional[str] = None,
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
+        evaluators: Optional[List[BaseEventEvaluator]] = None,
+        max_evaluator_concurrency: int = 5,
         prompt_tracking: Optional[Dict] = None,
     ) -> SendEventResponse:
```
nicolewhite, line 246: shouldn't this be gone now? it should be `-> None`

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +510,189 @@ def test_tracer_start_span(*args, **kwargs):
...
+def test_tracer_prod_async_evaluations(httpx_mock):
+    test_evaluation_id = uuid.uuid4()
+
+    class MyEvaluator1(BaseEventEvaluator):
+        id = "my-evaluator-1"
+
+        async def evaluate_event(self, event: TracerEvent) -> Evaluation:
+            return Evaluation(
+                evaluator_external_id=self.id,
+                id=test_evaluation_id,
```
nicolewhite, line 587: these should be dropped

nicolewhite (commented):
nice!! i think we just need to figure out async cleanup on program exit (SIGTERM, etc)

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,19 @@ def init() -> None:
...
 
 
+def _on_unexpected_exit(signum, frame):
+    if _loop and _loop.is_running():
+        log.info(f"Attempting to gracefully stop event loop. Received signal {signum}")
+        _loop.stop()
+        _loop.close()
+    exit(0)
+
+
```
dmorton2297, line 44: @nicolewhite my initial stab at attempting a graceful exit  - Struggling to come up with something to test this

nicolewhite: Where are we waiting for the futures to complete?

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,19 @@ def init() -> None:
         daemon=True,
     )
     background_thread.start()
-
+    signal.signal(signal.SIGTERM, _on_unexpected_exit)
+    signal.signal(signal.SIGINT, _on_unexpected_exit)
     _started = True
 
 
+def _on_unexpected_exit(signum, frame):
```
nicolewhite, line 37: I wouldn't necessarily call it unexpected, sigterm is pretty standard eg in Kubernetes

autoblocks/_impl/global_state.py
dmorton2297: Maybe something like this can be added

```python
def _on_exit(signum, frame):
    if _loop and _loop.is_running():
        log.info(f"Attempting to gracefully stop event loop. Received signal {signum}")
        _loop.stop()
        pending = asyncio.all_tasks()
        _loop.run_until_complete(asyncio.gather(*pending))
    exit(0)
```

Bit stuck on this part

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,20 @@ def init() -> None:
         daemon=True,
     )
     background_thread.start()
-
+    signal.signal(signal.SIGTERM, _on_exit)
+    signal.signal(signal.SIGINT, _on_exit)
     _started = True
 
 
+def _on_exit(signum, frame):
```
nicolewhite, line 37: can you add types

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,20 @@ def init() -> None:
...
+    if _loop and _loop.is_running():
+        log.info(f"Attempting to gracefully stop event loop. Received signal {signum}")
+        _loop.stop()
```
nicolewhite, line 40: is this right? we want to call stop before trying to clean up?

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,20 @@ def init() -> None:
...
+        pending = asyncio.all_tasks()
```
nicolewhite, line 41: can we do this only for the send event tasks? how does this work with 1) the main loop itself and 2) while loops like [this](https://github.com/autoblocksai/python-sdk/blob/175730e44294a56aa8a364cb7672c32a43934999/autoblocks/_impl/prompts/manager.py#L143)?

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,20 @@ def init() -> None:
...
+        _loop.run_until_complete(asyncio.gather(*pending))
```
nicolewhite, line 42: don't we want return_exceptions=True here?

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,20 @@ def init() -> None:
...
+    exit(0)
```
nicolewhite, line 43: can we avoid setting the exit code ourselves?

autoblocks/_impl/global_state.py
nicolewhite, line 41: can we log how many pending tasks there are

autoblocks/_impl/global_state.py
dmorton2297: Following this example: https://stackoverflow.com/questions/37417595/graceful-shutdown-of-asyncio-coroutines#:~:text=How%20can%20we%20wait%20tasks%20finished

Looking at this more closely though I think you are right

nicolewhite: hmm yeah from 2016, would be nice to find something more recent

autoblocks/_impl/global_state.py
dmorton2297: I actually think we can just straight omit it, since this would be the last block of code to run before exit

autoblocks/_impl/global_state.py
dmorton2297: added

autoblocks/_impl/global_state.py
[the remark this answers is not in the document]
dmorton2297: Looking into this now 👍

autoblocks/_impl/global_state.py
dmorton2297: For 2 - this code will not work. 

`asyncio.all_tasks` will get all the tasks from the global event loop - including the infinite loop task you created to refresh. 

At first glance I don't see an easy way to tell what the tasks are. 


What do you think about having a dedicated loop for your refresh action? That way I can specifically grab the main event loop which shouldn't contain any tasks that don't resolve

autoblocks/_impl/global_state.py
nicolewhite: as for testing, you could try putting together a script that has evaluators that just sleep for like 5 seconds and run that script with `subprocess`, kill the process, then check the  events exist by fetching them via the public api. i'd add it to tests/e2e

```
import subprocess

proc = subprocess.Popen(...)
proc.send_signal(signal.SIGTERM)

# wait for events to exist in public api
```

or something, haven't tried it

autoblocks/_impl/global_state.py
nicolewhite: i think the issue with that is it will be easy for us to forget the global event loop is off-limits for infinite loops. i think a better way is a dedicated loop for events, but then things are getting pretty complicated. is it possible to attach metadata to futures or otherwise inspect them? they're really just objects and python is the wild west, we might be able to do something to denote they are event futures

autoblocks/_impl/global_state.py
nicolewhite: i think we need this actually, when testing locally it never exits

autoblocks/_impl/global_state.py
dmorton2297: Mirrored approach here: https://www.roguelynn.com/words/asyncio-graceful-shutdowns/

Was able to use the coros to get tasks that were `_send_event_unsafe`

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,19 @@ def init() -> None:
         daemon=True,
     )
     background_thread.start()
-
+    for sig in [signal.SIGINT, signal.SIGTERM]:
+        _loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_on_exist_signal()))
     _started = True
 
 
+async def _on_exist_signal() -> None:
```
nicolewhite, line 37:
```suggestion
async def _on_exit_signal() -> None:
```

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,19 @@ def init() -> None:
...
+    tasks = [x for x in asyncio.all_tasks(loop=_loop) if x.get_coro().__name__ == "_send_event_unsafe" and not x.done()]
+    logging.info(f"Attempting to flush f{len(tasks)} outstanding send event tasks")
+    await asyncio.gather(*tasks)
```
nicolewhite, line 40: can you add back return_exceptions=True

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +29,19 @@ def init() -> None:
...
     )
     background_thread.start()
-
+    for sig in [signal.SIGINT, signal.SIGTERM]:
+        _loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_on_exist_signal()))
     _started = True
 
 
+async def _on_exist_signal() -> None:
+    tasks = [x for x in asyncio.all_tasks(loop=_loop) if x.get_coro().__name__ == "_send_event_unsafe" and not x.done()]
```
nicolewhite, line 38: can we add a test or something that will fail if we rename `_send_event_unsafe` to something else? as-is this is kind of out of sight and out of mind, so it would probably be helpful to have something that makes this relationship clear. so maybe:

`util.py`
```python
# Add long comment explaining why we do this
SEND_EVENT_CORO_NAME = "_send_event_unsafe"
```

in our tests:

```python
def test_tracer_has_send_event_unsafe():
    assert AutoblocksTracer._send_event_unsafe.__name__ == SEND_EVENT_CORO_NAME
```

so that if we come in later and forget we've done this and rename the function, this test will fail and we'll find the comment explaining this past decision

autoblocks/_impl/global_state.py
nicolewhite, line 37: also i think this function could use some comments, and maybe a link to the blog you followed

autoblocks/_impl/tracer.py
```diff
@@ -101,38 +109,120 @@ def start_span(self):
...
+                log.error(f"Unable to execute evaluator with id: {evaluator.id}. Error: %s", err, exc_info=True)
+
+        return evaluation
+
+    async def _run_evaluators(
+        self, evaluators: List[BaseEventEvaluator], event: TracerEvent, max_evaluator_concurrency: int
+    ) -> List[Dict[str, Any]]:
+        if len(evaluators) == 0:
+            return []
+        try:
```
nicolewhite, line 150: since you're already try-excepting up top you could probably drop this one and rename `_run_evaluators` to `_run_evaluators_unsafe` to follow the same pattern as send event. this basically lets you write a "clean" function w/o any error handling and then push the try-except to the outer call

autoblocks/_impl/tracer.py
```diff
@@ -150,8 +240,10 @@ def send_event(
         parent_span_id: Optional[str] = None,
         timestamp: Optional[str] = None,
         properties: Optional[Dict] = None,
+        evaluators: Optional[List[BaseEventEvaluator]] = None,
+        max_evaluator_concurrency: int = 5,
         prompt_tracking: Optional[Dict] = None,
-    ) -> SendEventResponse:
+    ) -> None:
```
nicolewhite, line 246: you need to drop line 274, should just return None in both cases

tests/autoblocks/test_tracer.py
```diff
@@ -30,12 +34,18 @@ def test_client_headers_init_with_key():
...
 
 
-@mock.patch.dict(
-    os.environ,
-    {
-        "AUTOBLOCKS_INGESTION_KEY": "mock-ingestion-key",
-    },
-)
+@pytest.fixture(autouse=True)
+def mock_evn_vars():
```
nicolewhite, line 38:
```suggestion
def mock_env_vars():
```

Blaso13 (commented):
Buuu no se podía vivir sin la rumana de las bicicletas @
IMG_20240107_183824
IMG_20240107_183814

autoblocks/_impl/global_state.py
dmorton2297: Added a test 👍

dmorton2297:
@nicolewhite Ok added the shutdown test here - this PR should be pretty close 🤞

autoblocks/_impl/util.py
```diff
@@ -51,4 +57,4 @@ async def sem_coro(coro: Coroutine[Any, Any, Any]) -> Any:
 
     # return_exceptions=True causes exceptions to be returned as values instead
     # of propagating them to the caller. this is similar in behavior to Promise.allSettled
-    await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
+    return await asyncio.gather(*(sem_coro(c) for c in coroutines), return_exceptions=True)
```
nicolewhite, line 60: i thought this `return await` didn't work, did something change?

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +441,215 @@ def test_tracer_start_span(*args, **kwargs):
...
+@mock.patch.object(
+    uuid,
+    "uuid4",
+    side_effect=["mock-uuid" for _ in range(2)],
+)
+def test_tracer_prod_async_evaluations(httpx_mock):
+    class MyEvaluator1(BaseEventEvaluator):
+        id = "my-evaluator-1"
+
+        async def evaluate_event(self, event: TracerEvent):
```
nicolewhite, line 505: should we add some `await asyncio.sleep`s here to catch the issues you were seeing with `return await`?

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +441,215 @@ def test_tracer_start_span(*args, **kwargs):
...
+        timestamp=timestamp,
+        properties={},
+        evaluators=[MyFailingEvaluator(), MyValidEvaluator()],
+    )
+
+
+@mock.patch.object(
+    AutoblocksTracer,
+    "_run_evaluators_unsafe",
+    side_effect=Exception("Something went wrong with our code when evaluating the event"),
```
nicolewhite, line 619: nice

tests/e2e/tracer_shutdown/main.py
```diff
@@ -0,0 +1,34 @@
...
+        return Evaluation(
+            score=0.9,
+            threshold=Threshold(gte=0.5),
+        )
+
+
+def main():
+    trace = sys.argv[1]
+    tracer.send_event("tracer_shutdown", trace_id=str(trace), evaluators=[MyEvaluator()])
+    time.sleep(10)  # Sleep for 10 seconds to allow time for signal to be sent
```
nicolewhite, line 30: i think we want this sleep in the evaluator so that the event isn't sent before it receives the signal

tests/e2e/tracer_shutdown/tracer_shutdown_test.py
```diff
@@ -0,0 +1,32 @@
...
+import time
+import unittest
+import uuid
+
+from autoblocks._impl.api.client import AutoblocksAPIClient
+
+client = AutoblocksAPIClient()
+
+
+class TestMainScript(unittest.TestCase):
```
nicolewhite, line 13: can you just add this to the existing tests/e2e file? also write it in pytest syntax

tests/e2e/tracer_shutdown/tracer_shutdown_test.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_tracer_received_sigint_or_sigterm(self):
+        test_trace_id = uuid.uuid4()
```
nicolewhite, line 15: wrap this in `str()` here so you don't have to repeat it twice

tests/e2e/tracer_shutdown/tracer_shutdown_test.py
```diff
@@ -0,0 +1,32 @@
...
+        # Start the main.py script as a subprocess
+        process = subprocess.Popen(["python", "tests/e2e/tracer_shutdown/main.py", str(test_trace_id)])
+        time.sleep(1)
```
nicolewhite, line 18: yeah since you're sleeping 1 here and only sleeping 1 in the evaluator it's likely the event was already sent before you even send the SIGTERM signal, which is why you need a longer sleep in your evaluator

tests/e2e/tracer_shutdown/tracer_shutdown_test.py
nicolewhite, line 1: in pytest the filenames need to be prefixed with `test_`

tests/e2e/tracer_shutdown/main.py
```diff
@@ -0,0 +1,34 @@
...
+
+tracer = AutoblocksTracer()
+
+
+# Send an event using the autoblocks tracer
+class MyEvaluator(BaseEventEvaluator):
+    id = "e2e-test-evaluator"
+
+    async def evaluate_event(self, event: TracerEvent) -> Evaluation:
+        await asyncio.sleep(1)
```
nicolewhite, line 20:
```suggestion
        await asyncio.sleep(10)
```

tests/e2e/tracer_shutdown/tracer_shutdown_test.py
```diff
@@ -0,0 +1,32 @@
...
+        # Wait for the process to terminate
+        process.wait(timeout=15)
+
+        test = client.get_trace(str(test_trace_id))
+        # Assert that the process has terminated successfully
+        self.assertIsNotNone(test)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
nicolewhite, line 32: you don't need this, `poetry run pytest` will automatically collect and run tests

nicolewhite (commented):
looks great! I think just need to switch to pytest syntax + rename the test file to match the others and we're good to go

autoblocks/_impl/util.py
dmorton2297: We were mislead - this wasn't working because we were mocking time in the unit test that this was getting hung up on. 


Issue as related to this: https://github.com/spulec/freezegun/pull/470

nicolewhite: ooh okay

tests/autoblocks/test_tracer.py
nicolewhite: oh right, gotcha

autoblocks/_impl/global_state.py
```diff
@@ -44,3 +73,9 @@ def http_client() -> httpx.AsyncClient:
     if not _client:
         raise Exception("HTTP client not initialized")
     return _client
+
+
+def sync_http_client() -> httpx.Client:
+    if not _sync_client:
+        raise Exception("HTTP client not initialized")
+    return _sync_client
```
dmorton2297, line 81: This is used in tracer - 

On shutdown - using an async http client does not behave nicely, as asyncio does not allow you use coroutines as the program is exiting. 

I figure using sync client is okay here, as the http call happens in the background thread and won't block execution

autoblocks/_impl/tracer.py
```diff
@@ -1,24 +1,55 @@
...
+@atexit.register
+def _cleanup_tracer() -> None:
+    """
+    On program exit, attempt to flush any outstanding send event tasks.
+    Since the background thread is a daemon, we have to actively
+    check task counts and sleep to measure progress.
+    """
+    num_tries = 0
+
+    def get_pending() -> List[asyncio.Task[Any]]:
```
dmorton2297, line 40: From the main thread - try to wait for tasks to wrap up that are running on the background daemon thread

dmorton2297:
Added to description - demo of graceful exit: https://drive.google.com/file/d/1ITe8iFdXY8qBAdUb3yBOPvyfNU9iJwkv/view?usp=sharing

autoblocks/_impl/tracer.py
```diff
@@ -1,24 +1,55 @@
...
+        return [
+            task
+            for task in asyncio.all_tasks(loop=global_state.event_loop())
+            if task.get_coro().__name__ == SEND_EVENT_CORO_NAME and not task.done()  # type: ignore
```
nicolewhite, line 44: can't we use the class directly to get the name now that it's in the same file?

autoblocks/_impl/tracer.py
```diff
@@ -1,24 +1,55 @@
...
+from autoblocks._impl.util import SEND_EVENT_CORO_NAME
 from autoblocks._impl.util import AutoblocksEnvVar
+from autoblocks._impl.util import gather_with_max_concurrency
 
 log = logging.getLogger(__name__)
 
 
-@dataclass
+@atexit.register
+def _cleanup_tracer() -> None:
```
nicolewhite, line 32: nit: you don't need underscores in this file because it's in the `_impl/` folder, so you don't need to worry about it being exposed to the end user

autoblocks/_impl/tracer.py
```diff
@@ -1,24 +1,55 @@
...
+    num_tries = 0
+
+    def get_pending() -> List[asyncio.Task[Any]]:
+        return [
+            task
+            for task in asyncio.all_tasks(loop=global_state.event_loop())
+            if task.get_coro().__name__ == SEND_EVENT_CORO_NAME and not task.done()  # type: ignore
+        ]
+
+    while len(get_pending()) > 0 and num_tries < 10:
```
nicolewhite, line 47: could we add some logs to this and also at the end

nicolewhite, line 47: 10 seconds is kinda short IMO, e.g. kubernetes default shutdown grace period is 30 seconds: https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace

tests/e2e/test_e2e.py
```diff
@@ -208,3 +211,19 @@ def test_prompt_manager_no_model_params_undeployed():
 
     with mgr.exec() as prompt:
         assert prompt.params is None
+
+
+def test_tracer_received_sigint_or_sigterm_cleanup():
```
nicolewhite, line 216: > test_tracer_received_sigint_or_sigterm_cleanup

it looks like this test is only testing SIGINT. can you add another one for SIGTERM? if you want you could use `pytest.parametrize `: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

tests/e2e/test_e2e.py
```diff
@@ -208,3 +211,19 @@ def test_prompt_manager_no_model_params_undeployed():
...
+    # Start the main.py script as a subprocess
+    process = subprocess.Popen(["python", "tests/e2e/tracer_shutdown_test_process.py", test_trace_id])
+    time.sleep(1)
+
+    os.kill(process.pid, signal.SIGINT)
+
+    # Wait for the process to terminate
+    process.wait()
+    time.sleep(10)  # give a moment for autoblocks to
+    test = client.get_trace(test_trace_id)
```
nicolewhite, line 227: can we do this in a while loop so we're not waiting a guaranteed 10 seconds

tests/e2e/test_e2e.py
```diff
@@ -208,3 +211,19 @@ def test_prompt_manager_no_model_params_undeployed():
...
+    test_trace_id = str(uuid.uuid4())
+    # Start the main.py script as a subprocess
+    process = subprocess.Popen(["python", "tests/e2e/tracer_shutdown_test_process.py", test_trace_id])
+    time.sleep(1)
+
+    os.kill(process.pid, signal.SIGINT)
+
+    # Wait for the process to terminate
+    process.wait()
+    time.sleep(10)  # give a moment for autoblocks to
```
nicolewhite, line 226: finish comment

autoblocks/_impl/tracer.py
dmorton2297: I think we can bump this up to 30 as that would be worse case 👍 In most scenarios I am hoping an immediate exit just happens (no events on queue)

tests/e2e/test_e2e.py
```diff
@@ -208,3 +211,19 @@ def test_prompt_manager_no_model_params_undeployed():
...
+    test = client.get_trace(test_trace_id)
+    # Assert that the process has terminated successfully
```
nicolewhite, line 228: nit: "and that the event was sent after receiving the shutdown signal"

tests/e2e/test_e2e.py
```diff
@@ -208,3 +211,19 @@ def test_prompt_manager_no_model_params_undeployed():
 
     with mgr.exec() as prompt:
         assert prompt.params is None
+
+
+def test_tracer_received_sigint_or_sigterm_cleanup():
+    test_trace_id = str(uuid.uuid4())
+    # Start the main.py script as a subprocess
```
nicolewhite, line 218: nit: this comment is referring to `main.py` but that is not the name of the script

autoblocks/_impl/global_state.py
nicolewhite: doesn't this defeat the purpose of using the async event loop?

autoblocks/_impl/global_state.py
```diff
@@ -25,10 +33,31 @@ def init() -> None:
         daemon=True,
     )
     background_thread.start()
-
+    for sig in [signal.SIGINT, signal.SIGTERM]:
+        _loop.add_signal_handler(sig, lambda: asyncio.create_task(_on_exit_signal()))
     _started = True
 
 
+async def _on_exit_signal() -> None:
```
nicolewhite, line 41: do we need both this + the atexit handler in tracer.py?

autoblocks/_impl/tracer.py
```diff
@@ -103,44 +134,119 @@ def start_span(self) -> Generator[None, None, None]:
...
         *,
         message: str,
-        trace_id: Optional[str] = None,
-        span_id: Optional[str] = None,
-        parent_span_id: Optional[str] = None,
-        timestamp: Optional[str] = None,
-        properties: Optional[Dict[Any, Any]] = None,
-        prompt_tracking: Optional[Dict[str, Any]] = None,
-    ) -> SendEventResponse:
+        max_evaluator_concurrency: int,
```
nicolewhite, line 196: can we move this arg to the bottom next to `evaluators` like in the other signature

autoblocks/_impl/tracer.py
nicolewhite: nvm, didn't see we're also waiting for events back in global_state...we should find a way to consolidate

autoblocks/_impl/tracer.py
```diff
@@ -103,44 +134,119 @@ def start_span(self) -> Generator[None, None, None]:
...
+                    event=event,
+                    evaluator=evaluator,
+                )
+                for evaluator in evaluators
+            ],
+        )
+        return [
+            self._evaluation_to_json(evaluation=evaluation, evaluator_external_id=evaluator.id)
+            for evaluator, evaluation in zip(evaluators, evaluations)
+            if evaluation is not None
```
nicolewhite, line 188: this actually needs to be `if isinstance(evaluation, Evaluation)` since it could be an instance of an exception

nicolewhite: noticed while testing something else

dmorton2297: Updated

autoblocks/_impl/tracer.py
dmorton2297: Updated

autoblocks/_impl/global_state.py
dmorton2297: Removed at exit / signal handler stuff for now

autoblocks/_impl/tracer.py
```diff
@@ -103,44 +110,119 @@ def start_span(self) -> Generator[None, None, None]:
...
+            except Exception as err:
+                log.error("Unable to evaluate events. Error: %s", err, exc_info=True)
+
+        traced_event = TracerEvent(
+            message=message,
+            trace_id=trace_id,
+            timestamp=timestamp,
+            properties=merged_properties,
+        )
+        req = global_state.sync_http_client().post(
```
nicolewhite, line 219: we need to switch back to the async client + drop sync_http_client

autoblocks/_impl/util.py
```diff
@@ -10,6 +10,11 @@
 
 log = logging.getLogger(__name__)
 
+# This constant is used during the shutdown process to find all outstanding
+# send event tasks and attempt to resolved them before stopping the event loop.
+# We have a test asserting that this function name does not change.
+SEND_EVENT_CORO_NAME = "_send_event_unsafe"
```
nicolewhite, line 16: can drop this

tests/autoblocks/test_tracer.py
```diff
@@ -448,3 +441,215 @@ def test_tracer_start_span(*args, **kwargs):
...
+    tracer.send_event(
+        "my-message",
+        trace_id="my-trace-id",
+        timestamp=timestamp,
+        properties={},
+        evaluators=[MyEvaluator()],
+    )
+
+
+def test_tracer_has_send_event_unsafe():
```
nicolewhite, line 654: drop

autoblocks/_impl/util.py
```diff
@@ -22,6 +27,7 @@ class AutoblocksEnvVar(StrEnum):
     CLI_SERVER_ADDRESS = "AUTOBLOCKS_CLI_SERVER_ADDRESS"
     INGESTION_KEY = "AUTOBLOCKS_INGESTION_KEY"
     TRACER_THROW_ON_ERROR = "AUTOBLOCKS_TRACER_THROW_ON_ERROR"
+    TRACER_BLOCK_ON_SEND_EVENT = "TRACER_BLOCK_ON_SEND_EVENT"
```
nicolewhite, line 30: drop

tests/autoblocks/test_tracer.py
```diff
@@ -30,12 +35,18 @@ def test_client_headers_init_with_key():
...
-        "AUTOBLOCKS_INGESTION_KEY": "mock-ingestion-key",
-    },
-)
+@pytest.fixture(autouse=True)
+def mock_env_vars():
+    with mock.patch.dict(
+        os.environ,
+        {
+            "AUTOBLOCKS_INGESTION_KEY": "mock-ingestion-key",
+            "TRACER_BLOCK_ON_SEND_EVENT": "1",
```
nicolewhite, line 44: drop

autoblocks/_impl/util.py
nicolewhite: also this should be prefixed with `AUTOBLOCKS_` like all the rest

nicolewhite (approved):
LGTM

PR #125, opened 2024-02-27
Add mypy

nicolewhite:
most of this is just adding `Any` to places where we don't know / care what the type is exactly. will need to do more research on what the recommended alternatives are to `Any` (like does python have an equivalent of unknown?) so that we can disallow Any like we do in our TS repos

running with `--strict` which will run with these options:

```
  --strict                  Strict mode; enables the following flags: --warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any, --disallow-untyped-calls, --disallow-untyped-defs,
                            --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-decorators, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-reexport,
                            --strict-equality, --extra-checks
```


.pre-commit-config.yaml
```diff
@@ -20,3 +20,13 @@ repos:
     hooks:
       - id: black
         language_version: python3.11
+
+  - repo: local
```
nicolewhite, line 24: i tried using the mypy pre-commit mirror but was having issues with the type stubs we need (like types-orjson, types-PyYaml, etc) so just running it directly

autoblocks/_impl/prompts/manager.py
```diff
@@ -197,7 +198,7 @@ def exec(self) -> ContextManager[ExecutionContextType]:
         """
 
         @contextlib.contextmanager
-        def gen():
+        def gen():  # type: ignore
```
nicolewhite, line 201: the `exec` above is already incorrectly typed due to a pycharm bug, so just ignoring this inner context manager that it's wrapping

autoblocks/_impl/prompts/models.py
```diff
@@ -80,7 +80,7 @@ def str_version(self) -> str:
 
 
 class PromptMinorVersion(FrozenModel):
-    version: Union[Enum, List[WeightedMinorVersion]]
+    version: Union[Enum, List[WeightedMinorVersion]]  # type: ignore
```
nicolewhite, line 83: need to do more research here, `WeightedMinorVersion` requires generics but we don't care what it is in this case, just that it's an instance of `WeightedMinorVersion`. tried wrapping in `Type` or using `Any` for the generics but couldn't get it working

autoblocks/_impl/prompts/models.py
```diff
@@ -89,7 +89,7 @@ def str_version(self) -> Optional[str]:
         return str(self.version.value)
 
     @property
-    def weighted_versions(self) -> Optional[List[WeightedMinorVersion]]:
+    def weighted_versions(self) -> Optional[List[WeightedMinorVersion]]:  # type: ignore
```
nicolewhite, line 92: same issue here

.github/workflows/ci.yml
```diff
@@ -47,7 +57,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.8"
```
nicolewhite, line 50: dropping 3.8 like we discussed

.github/workflows/ci.yml
```diff
@@ -34,6 +34,16 @@ jobs:
...
+      - name: Install poetry
+        run: curl -sSL https://install.python-poetry.org | python3 -
+
+      - name: Check pyproject.toml & poetry.lock are in sync
+        run: poetry lock --check
+
+      # We only need to install the dev dependencies, which will install mypy and related packages
+      - name: Install dev dependencies
+        run: poetry install --only=dev
+
```
dmorton2297, line 46: Just an understanding check

We have to install deps in our CI because of the local pre-commit hook you are defining below, which depends on `mypy` and others

autoblocks/_impl/testing/run.py
```diff
@@ -59,8 +58,8 @@ async def send_error(
     evaluator_id: Optional[str],
     error: Exception,
 ) -> None:
-    await client.post(
-        "/errors",
+    await global_state.http_client().post(
+        f"{cli()}/errors",
```
dmorton2297, line 62: 👍

tests/autoblocks/test_prompts_autogenerate.py
```diff
@@ -61,7 +61,7 @@ def test_infer_type():
     assert infer_type(0.0) == "Union[float, int]"
     assert infer_type(True) == "bool"
     assert infer_type(False) == "bool"
-    assert infer_type([""]) == "List[str]"
+    assert infer_type([""]) == "list[str]"
```
dmorton2297, line 64: Another understanding check - 

We were using `types.List` before because the built in list didn't support subscriping (like list[str]) until python 3.9 - now that we are dropping 3.9, we can just use `list`

dmorton2297 (dismissed):

tests/autoblocks/test_prompts_autogenerate.py
nicolewhite: yeah exactly

.github/workflows/ci.yml
nicolewhite: yep

dmorton2297 (dismissed):

adamnolte (approved):

PR #128, opened 2024-02-29
Move max_concurrency to evaluator classes

nicolewhite:


tests/autoblocks/test_run_test_suite.py
```diff
@@ -604,6 +604,224 @@ def evaluate_test_case(self, test_case: MyTestCase, output: str) -> Evaluation:
...
+        evaluators=[
+            EvaluatorA(),
+            EvaluatorB(),
+        ],
+        fn=test_fn,
+        max_test_case_concurrency=1,
+        max_evaluator_concurrency={
+            "evaluator-a": 1,
+            "evaluator-b": 1,
+        },
```
nicolewhite, line 640: @adamnolte this is what i'm thinking for per-evaluator concurrency

autoblocks/_impl/testing/run.py
```diff
@@ -230,7 +248,6 @@ async def async_run_test_suite(
                     test_case=test_case,
                     evaluators=evaluators,
                     fn=fn,
-                    max_evaluator_concurrency=max_evaluator_concurrency,
```
nicolewhite, line 233: i think the way we're doing test case concurrency could also use some tweaking. right now a test case is in considered "in progress" for both the duration of running it thru `fn` and running all of its evaluators, but we probably only need to limit concurrency for the task of running it thru `fn`

tests/autoblocks/test_run_test_suite.py
adamnolte: Can we statically reference the id's on the evaluators? So like EvaluatorA.id ?

nicolewhite: yeah good call

nicolewhite: we could also make it a property of the evaluators themselves:

```python
class EvaluatorA(BaseTestEvaluator):
    max_concurrency = 5
```

nicolewhite: that would probably make more sense since it would be awkward to have to repeat yourself across multiple test suites if you're sharing an evaluator

adamnolte: ya was thinking about that, I think i actually prefer that. @dmorton2297 wdyt?

autoblocks/_impl/testing/run.py
```diff
@@ -73,43 +81,52 @@ async def send_error(
...
-                test_case_hash=test_case._cached_hash,
-                evaluator_id=evaluator.id,
-                error=err,
-            )
-    else:
-        try:
+) -> Evaluation:
+    """
+    This is suffixed with _unsafe because doesn't handle exceptions.
+    Its caller will catch and handle all exceptions.
```
dmorton2297, line 92:
```suggestion
    This is suffixed with _unsafe because it doesn't handle exceptions.
    Its caller will catch and handle all exceptions.
```

autoblocks/_impl/testing/run.py
```diff
@@ -52,6 +52,14 @@ def serialize(x: Any) -> Any:
     return orjson.loads(orjson.dumps(x, default=orjson_default))
 
 
+async def all_settled(coroutines: list[Coroutine[Any, Any, Any]]) -> None:
+    """
+    Runs all the coroutines in parallel and waits for all of them to finish,
+    regardless of whether they succeed or fail. Similar to Promise.allSettled.
+    """
+    await asyncio.gather(*coroutines, return_exceptions=True)
+
```
dmorton2297, line 61: Could we potentially reuse `gather_with_max_concurrency` from `util.py` or consolidate?

nicolewhite: this doesn't have a max concurrency and i don't think we should modify `gather_with_max_concurrency` to handle situations w/ no max concurrency, since that is all it does

nicolewhite: i'm also going to delete it anyway when both your PR and https://github.com/autoblocksai/python-sdk/pull/128 are merged b/c we'll no longer have the `max_evaluator_concurrency` arg

nicolewhite:
@dmorton2297 @adamnolte updated this to handle concurrency in the online evals as well

adamnolte (approved):

PR #131, opened 2024-03-04
App graceful shutdown

nicolewhite:


autoblocks/_impl/global_state.py
```diff
@@ -11,58 +11,60 @@
...
 _client: Optional[httpx.AsyncClient] = None
-_sync_client: Optional[httpx.Client] = None
 _loop: Optional[asyncio.AbstractEventLoop] = None
+_background_thread: Optional[threading.Thread] = None
 _started: bool = False
 
 
 def init() -> None:
-    global _client, _loop, _started, _sync_client
+    global _client, _loop, _started, _background_thread
```
dmorton2297, line 20: Why global? I don't see this being used in tracer - my initial thought was to make this global and do a `global_state.background_thread().join` in the atexit handler of tracer

nicolewhite: oops good catch, was leftover from something else i had tried

PR #135, opened 2024-03-08
Fix some errors not propagating

nicolewhite:
need to fix in the TS SDK too

### DIFF
--- a/autoblocks/_impl/testing/run.py
+++ b/autoblocks/_impl/testing/run.py
@@ -9,12 +9,11 @@ from typing import Callable
 from typing import List
 from typing import Optional
 
-import orjson
-
 from autoblocks._impl import global_state
 from autoblocks._impl.testing.models import BaseTestCase
 from autoblocks._impl.testing.models import BaseTestEvaluator
-from autoblocks._impl.testing.models import Evaluation
+from autoblocks._impl.testing.util import serialize
+from autoblocks._impl.testing.util import serialize_test_case
 from autoblocks._impl.util import AutoblocksEnvVar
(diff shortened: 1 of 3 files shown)


autoblocks/_impl/testing/run.py
```diff
@@ -38,36 +37,6 @@ def cli() -> str:
     return cli_server_address
 
 
-def orjson_default(o: Any) -> Any:
```
nicolewhite, line 41: unrelated, just moving these into `util`

autoblocks/_impl/testing/run.py
```diff
@@ -161,10 +127,22 @@ async def run_test_case_unsafe(
...
+            output = await global_state.event_loop().run_in_executor(None, ctx.run, fn, test_case)
+
+    await global_state.http_client().post(
+        f"{cli()}/results",
+        json=dict(
+            testExternalId=test_id,
+            testCaseHash=test_case._cached_hash,
+            testCaseBody=serialize_test_case(test_case),
+            testCaseOutput=serialize(output),
+        ),
```
nicolewhite, line 142: the issue is that this part was not in the `_unsafe` method but was being called afterward where we have no outer error handler sending errors, so this was being swallowed by `all_settled`, where we don't do anything w/ the results

note that we're not try-catching the POST request here, since the client already suppresses those (those errors are handled and displayed by the CLI). here we are making sure we are try-catching the building of the JSON payload, which is where the error was occurring

adamnolte (approved):

nicolewhite:
@adamnolte I believe you suggested I do this in a past review as well and I decided not to 😭

PR #136, opened 2024-03-09
Send tracer events to cli in test mode

adamnolte:
Synchronously sends events to the cli when we detect a cli environment variable. The cli just collects these into an array, so doing this synchronously will have minimal impact on the test run.

Uses context variables to associate test id and test case hash. The test is setup so that it verifies when we have multiple test cases running at a time, everything is working correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
 
   tests:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
     strategy:
       fail-fast: false
--- a/autoblocks/_impl/testing/run.py
+++ b/autoblocks/_impl/testing/run.py
@@ -127,7 +127,6 @@ async def run_test_case_unsafe(
     This is suffixed with _unsafe because it doesn't handle exceptions.
     Its caller will catch and handle all exceptions.
     """
-    token = current_test_case_hash.set(test_case._cached_hash)
     async with test_case_semaphore_registry[test_id]:
 
         if inspect.iscoroutinefunction(fn):
@@ -145,7 +144,6 @@ async def run_test_case_unsafe(
             testCaseOutput=serialize(output),
         ),
     )
-    current_test_case_hash.reset(token)
     return output
 
 
@@ -155,6 +153,7 @@ async def run_test_case(
     evaluators: List[BaseTestEvaluator],
     fn: Callable[[BaseTestCase], Any],
 ) -> None:
+    token = current_test_case_hash.set(test_case._cached_hash)
     try:
         output = await run_test_case_unsafe(
             test_id=test_id,
@@ -169,6 +168,8 @@ async def run_test_case(
             error=err,
         )
         return
+    finally:
+        current_test_case_hash.reset(token)
 
     try:
         await all_settled(
@@ -259,9 +260,10 @@ async def async_run_test_suite(
             evaluator_id=None,
             error=err,
         )
+    finally:
+        current_external_test_id.reset(token)
 
     await global_state.http_client().post(f"{cli()}/end", json=dict(testExternalId=test_id))
-    current_external_test_id.reset(token)
 
 
 def run_test_suite(
--- a/autoblocks/_impl/testing/run.py
+++ b/autoblocks/_impl/testing/run.py
@@ -127,12 +127,14 @@ async def run_test_case_unsafe(
     This is suffixed with _unsafe because it doesn't handle exceptions.
     Its caller will catch and handle all exceptions.
     """
+    token = current_test_case_hash.set(test_case._cached_hash)
     async with test_case_semaphore_registry[test_id]:
-        current_test_case_hash.set(test_case._cached_hash)
+
         if inspect.iscoroutinefunction(fn):
             output = await fn(test_case)
         else:
             ctx = contextvars.copy_context()
+            print(ctx.items())
             output = await global_state.event_loop().run_in_executor(None, ctx.run, fn, test_case)
 
     await global_state.http_client().post(
@@ -144,7 +146,7 @@ async def run_test_case_unsafe(
             testCaseOutput=serialize(output),
         ),
     )
-
+    current_test_case_hash.reset(token)
     return output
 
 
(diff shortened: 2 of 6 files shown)


autoblocks/_impl/testing/run.py
```diff
@@ -141,7 +145,7 @@ async def run_test_case_unsafe(
             testCaseOutput=serialize(output),
         ),
     )
-
+    current_test_case_hash.reset(token)
```
adamnolte, line 148: this isn't necessary, but more of a safety mechanism

autoblocks/_impl/testing/run.py
```diff
@@ -256,6 +261,7 @@ async def async_run_test_suite(
         )
 
     await global_state.http_client().post(f"{cli()}/end", json=dict(testExternalId=test_id))
+    current_external_test_id.reset(token)
```
adamnolte, line 264: this isn't necessary, but more of a safety mechanism

.github/workflows/ci.yml
```diff
@@ -52,6 +52,7 @@ jobs:
 
   tests:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
```
adamnolte, line 55: i had an issue with freezegun and async tests so added this to prevent ci from just running forever when the tests hang

adamnolte:
I did also test this in a python project (user feedback clustering)

autoblocks/_impl/tracer.py
```diff
@@ -244,6 +231,45 @@ async def _send_event_unsafe(
...
+            log.error("Failed to send test event to Autoblocks. No test case hash set.")
+            return
+
+        traced_event = TracerEvent(
+            message=message,
+            trace_id=trace_id,
+            timestamp=timestamp,
+            properties=properties,
+        )
+        req = global_state.sync_http_client().post(
```
nicolewhite, line 262: Would prefer we just use the async client for everything to be consistent but I guess this works if we're doing it on the main thread and not in our event loop

autoblocks/_impl/tracer.py
```diff
@@ -264,16 +290,38 @@ def send_event(
         the event failed, the trace_id will be None.
         """
         try:
+            merged_properties = dict(self._properties)
```
nicolewhite, line 293: Would be nice to pull all of this property merging stuff into a function just bc I think functions with a large amount of lines in a try/except are hard to follow

nicolewhite (dismissed):
Left some thoughts but not blockers

autoblocks/_impl/testing/run.py
nicolewhite: Do we want to reset to the previous value? Or just set it to None

adamnolte: The previous value would be none here

autoblocks/_impl/tracer.py
nicolewhite: Oh wait I guess that wouldn't even work right? It was blocking?

autoblocks/_impl/testing/run.py
nicolewhite: Oh right ok

autoblocks/_impl/tracer.py
adamnolte: Well this function is running in our event loop since it's in the test context. I think we'll always want this to be blocking in the test context

nicolewhite: Oh well in that case it would be nice to use the async client because when you use a sync client you're blocking the loop until it completes instead of yielding to other tasks

adamnolte: Since it's just sending to the local cli server (and that just adds to an in memory array and returns) it should be very fast.

I'd have to use event-loop.execute to use async right? I can see if it can get that work

nicolewhite: I think in the testing context it doesn't matter too much what we do here. And I think we're going to switch to queue + sync anyway right? So maybe we can leave it

adamnolte: Ya I was just opting for simple in test context. I'm not sure if we'll want queue for test events. We'd have to make sure to flush it before we send test case results

nicolewhite: Ya I meant for prod

autoblocks/_impl/testing/run.py
nicolewhite: it's also possible this wouldn't be reached if there is an uncaught error in `fn`, you'd have to put this in a `finally` block

autoblocks/_impl/tracer.py
```diff
@@ -244,6 +231,66 @@ async def _send_event_unsafe(
...
+            json=dict(
+                testExternalId=current_external_test_id.get(),
+                testCaseHash=current_test_case_hash.get(),
+                event=traced_event.to_json(),
+            ),
+            timeout=self._timeout_seconds,
+        )
+        req.raise_for_status()
+
+    def merge_properties(
```
nicolewhite: can you add a leading underscore since this is an internal function

adamnolte: ya i already did if you refresh 😄

autoblocks/_impl/testing/run.py
adamnolte: hmm ya good call, let me move it up a level

nicolewhite: i don't think it matters too much since it's unnecessary like you said, was just a thought

adamnolte: ya it doesn't matter for this use case, but good hygeine in case we use more and people reference this, I did the same for the test id too.

autoblocks/_impl/testing/run.py
```diff
@@ -234,7 +239,7 @@ async def async_run_test_suite(
     }
 
     await global_state.http_client().post(f"{cli()}/start", json=dict(testExternalId=test_id))
-
```
nicolewhite, line 237: would be nice to keep the original spacing

autoblocks/_impl/testing/run.py
```diff
@@ -126,6 +128,7 @@ async def run_test_case_unsafe(
     Its caller will catch and handle all exceptions.
     """
     async with test_case_semaphore_registry[test_id]:
+
```
nicolewhite, line 131: same here would be nice to drop this added space and the removed one below

nicolewhite (dismissed):

autoblocks/_impl/testing/run.py
nicolewhite: ok nice

nicolewhite (approved):

PR #142, opened 2024-03-14
Make checking for test run context consistent with JS SDK

adamnolte:
This will make it so even if you are running through our cli, if you use send event outside of a test run events will still send. This will be consistent with - https://github.com/autoblocksai/javascript-sdk/pull/120

### DIFF
--- a/autoblocks/_impl/context_vars.py
+++ b/autoblocks/_impl/context_vars.py
@@ -10,5 +10,5 @@ class TestCaseRunContext:
 
 
 test_case_run_context_var: ContextVar[Optional[TestCaseRunContext]] = ContextVar(
-    "autoblocks_sdk_current_external_test_id", default=None
+    "autoblocks_sdk_test_case_run_context_var", default=None
 )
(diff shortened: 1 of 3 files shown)


autoblocks/_impl/context_vars.py
```diff
@@ -1,7 +1,14 @@
...
-current_external_test_id: ContextVar[Optional[str]] = ContextVar(
+
+@dataclasses.dataclass
+class TestCaseRunContext:
+    test_id: str
+    test_case_hash: str
+
+
+test_case_run_context_var: ContextVar[Optional[TestCaseRunContext]] = ContextVar(
     "autoblocks_sdk_current_external_test_id", default=None
```
nicolewhite, line 13: can we rename the context var here now that it has both

autoblocks/_impl/testing/run.py
```diff
@@ -266,8 +265,6 @@ async def async_run_test_suite(
             evaluator_id=None,
             error=err,
         )
-    finally:
-        current_external_test_id.reset(token)
```
nicolewhite, line 270: you don't wanna do this anymore?

adamnolte: this happens in the test case run itself now since I combined them.

nicolewhite: oh ya i see

nicolewhite (approved):

PR #144, opened 2024-03-14
handle send_event while already in event loop

nicolewhite:


autoblocks/_impl/tracer.py
```diff
@@ -27,6 +31,28 @@
 
 evaluator_semaphore_registry: dict[str, asyncio.Semaphore] = {}  # evaluator_id -> semaphore
 
+background_tasks: Set[Union[asyncio.Task[Any], Future[Any]]] = set()
+
+
+@atexit.register
```
nicolewhite, line 37: this is not doing anything since we're still calling `result()` so should always be empty on exit in theory, but wondering if this approach will work in general (as opposed to a queue)

tests/autoblocks/test_tracer.py
```diff
@@ -50,6 +51,19 @@ def test_client_init_headers_with_env_var():
     assert tracer._client_headers["Authorization"] == "Bearer mock-ingestion-key"
 
 
+def expect_post_request(
```
nicolewhite, line 54: i will update the other tests to use this in another PR

adamnolte (approved):

autoblocks/_impl/tracer.py
adamnolte: ya i was wondering the same, wasn't sure exactly all you and dan had tried in the other approach.

nicolewhite: I think we got stuck on keeping the daemon thread alive on shutdown iirc, might run into the same issue here

PR #145, opened 2024-03-15
misc prompt sdk cleanup

nicolewhite:
was reading thru and wanted to do a bit of cleanup before adding the prompt snapshot stuff


adamnolte (dismissed):

nicolewhite:
@adamnolte sorry one more change, decided to just send the params as-is so that both our `PromptTracking` schema and the schema we return from the REST API are the same. also updated the JS side: https://github.com/autoblocksai/javascript-sdk/pull/122

adamnolte (approved):

PR #148, opened 2024-03-18
Add BaseEvaluator

nicolewhite:
to mirror https://github.com/autoblocksai/javascript-sdk/pull/123


autoblocks/testing/models.py
```diff
@@ -1,3 +1,4 @@
+from autoblocks._impl.testing.models import BaseEvaluator  # noqa: F401
```
nicolewhite, line 1: @adamnolte i will tackle "exporting" these from an `evaluators` module as a separate task since we'll need to be careful to leave the old exports for backwards compatability

dmorton2297 (approved):
Looking at JS pr now to catch up

PR #149, opened 2024-03-18
EPD-695 Use snapshot if it is set

nicolewhite:


autoblocks/_impl/prompts/manager.py
```diff
@@ -32,6 +35,32 @@
...
 
+def is_testing_context() -> bool:
+    """
+    Note that we check for the presence of the CLI environment
+    variable and not the test case contextvars because the
+    contextvars aren't set until run_test_suite is called,
+    whereas a prompt manager might have already been imported
+    and initialized by the time run_test_suite is called.
+    """
+    return bool(AutoblocksEnvVar.CLI_SERVER_ADDRESS.get())
```
adamnolte, line 46: Should we just check for the prompt snapshots env variable?

autoblocks/_impl/prompts/manager.py
```diff
@@ -111,18 +156,61 @@ async def _get_prompt(
...
+        )
+
+        if resp.status_code == HTTPStatus.CONFLICT:
+            # The endpoint returns this status code when the snapshot is
+            # not compatible with the major version this prompt manager
+            # is configured to use.
+            raise IncompatiblePromptSnapshotError(
+                f"Can't override '{self._name}' with prompt snapshot '{snapshot_id}' because it is not compatible "
+                f"with major version '{self.__prompt_major_version__}'."
+            )
```
adamnolte, line 186: i'm wondering how we can easily surface this in the UI, most likely the people that triggered this error aren't going to look at the logs of the run to know what happened.

nicolewhite: yeah really good point, i have a few ideas. i'll create a ticket and put thoughts there

autoblocks/_impl/prompts/manager.py
nicolewhite: i'm using this to know if we're in a testing context at all (even if prompt snapshots are not set) to disable the latest refresh, since i don't think it makes sense in a testing context since you probably want the prompt to be stable during a test run

nicolewhite: https://github.com/autoblocksai/python-sdk/pull/149/files#diff-4e124f39d32018a40189c28c9452a15f042fd1c6eb54d58eeb286ed42c8b6381R122

adamnolte: ahh i see

adamnolte (approved):

PR #153, opened 2024-03-26
EPD-758 Send events in background

nicolewhite:


autoblocks/_impl/global_state.py
```diff
@@ -1,20 +1,105 @@
...
+    """
+    main_thread = threading.main_thread()
+    main_thread.join()
+
+    # Used by the tracer to switch to the sync HTTP client
+    # after the interpreter has shut down. Otherwise we'll get:
+    # RuntimeError: cannot schedule new futures after interpreter shutdown
+    # TODO: figure out how to keep the main thread alive while
+    # we flush the tasks so that we don't need to switch to the
+    # sync client. Users will also run into issues if their
```
dmorton2297, line 92: Just a note to myself: Conceptually - All we should have to do is not use a daemon thread for the background worker - because in that case the main thread won't exit until all other threads are joined.

I went down this path - We both ended up at the same error though (interpreter shutdown)

autoblocks/_impl/global_state.py
```diff
@@ -1,20 +1,105 @@
...
+from typing import Set
 
 import httpx
 
+from autoblocks._impl.util import AnyTask
+
 log = logging.getLogger(__name__)
 
+_started: bool = False
+_thread: Optional[threading.Thread] = None
```
dmorton2297, line 16: Maybe name `_background_worker_thread` or something like that to be explicit

autoblocks/_impl/tracer.py
```diff
@@ -238,13 +214,31 @@ async def _send_event_unsafe(
...
 
-        req = await global_state.http_client().post(
-            url=INGESTION_ENDPOINT,
-            json=payload.to_json(),
-            headers=self._client_headers,
-            timeout=self._timeout_seconds,
-        )
-        req.raise_for_status()
+        try:
+            if global_state.main_thread_has_finished():
```
dmorton2297, line 218: I would just add one comment about how this is an edge case for normal shutdown

tests/e2e/plain_script.py
```diff
@@ -0,0 +1,31 @@
...
+from tests.e2e.slow_evaluators import SlowEvaluator1
+from tests.e2e.slow_evaluators import SlowEvaluator2
+
+log = logging.getLogger(__name__)
+
+tracer = AutoblocksTracer()
+
+
+if __name__ == "__main__":
+    logging.basicConfig(
```
dmorton2297, line 14: 🔥

autoblocks/_impl/tracer.py
dmorton2297, line 218: I would just add a small comment about how this should only happen in the normal shutdown scenario

nicolewhite: it will happen for both plain script and sigterm because we are not starting the flush until the main thread has joined

autoblocks/_impl/global_state.py
nicolewhite: > because in that case the main thread won't exit until all other threads are joined

are you sure? i think we wouldn't need the workaround if this were true, because then we would not see the "cannot schedule new futures" error while our non-daemon background thread is still running

dmorton2297: These are the docs that made me think this

"
A thread can be flagged as a “daemon thread”. The significance of this flag is that the entire Python program exits when only daemon threads are left. The initial value is inherited from the creating thread. The flag can be set through the [daemon](https://docs.python.org/3/library/threading.html#threading.Thread.daemon) property or the daemon constructor argument.

Note Daemon threads are abruptly stopped at shutdown. Their resources (such as open files, database transactions, etc.) may not be released properly. If you want your threads to stop gracefully, make them non-daemonic and use a suitable signalling mechanism such as an [Event](https://docs.python.org/3/library/threading.html#threading.Event).
"

I think the issue is how `main` spawns the interpreter (as defined in https://docs.python.org/3/library/threading.html#threading.main_thread) that other threads use?

dmorton2297 (approved):

autoblocks/_impl/global_state.py
nicolewhite: i think it's actually a bug that started in python 3.9 wrt ThreadPoolExecutor: https://bugs.python.org/issue42647

PR #155, opened 2024-03-29
Add helper to get test cases

dmorton2297:
Added new function to http client that will allow a user to fetch test cases given a testId


autoblocks/_impl/api/models.py
```diff
@@ -46,6 +46,13 @@ class DatasetWithItems(Dataset):
     items: List[DatasetItem]
 
 
+@dataclass
+class ManagedTestCase:
+    body: Dict[str, Any]
+    createdAt: str
+    updatedAt: str
```
nicolewhite, line 53: python is snake_case

but let's just drop these both here and in the endpoint since they're not necessary

tests/autoblocks/test_api_client.py
```diff
@@ -35,6 +35,20 @@ def test_client_init_with_env_var():
     assert client._client.headers.get("authorization") == "Bearer mock-api-key"
 
 
+def test_get_test_cases(httpx_mock):
```
nicolewhite, line 38: let's also add an e2e test

autoblocks/_impl/api/models.py
dmorton2297: These changes aren't ready for review yet - will update

adamnolte (approved):

PR #161, opened 2024-04-04
update types to pass mypy --strict

nicolewhite:
if you run `mypy --strict` on the testing-sdk project in the examples repo you get a ton of errors, this fixes all of them

### DIFF
--- a/autoblocks/_impl/testing/models.py
+++ b/autoblocks/_impl/testing/models.py
@@ -2,8 +2,12 @@ import abc
 import dataclasses
 import functools
 from typing import Any
+from typing import Awaitable
 from typing import Dict
+from typing import Generic
 from typing import Optional
+from typing import TypeVar
+from typing import Union
 
 
 @dataclasses.dataclass
@@ -44,15 +48,19 @@ class BaseTestCase(abc.ABC):
         pass
 
 
+TestCaseType = TypeVar("TestCaseType", bound=BaseTestCase)
+OutputType = TypeVar("OutputType")
+
+
 @dataclasses.dataclass
-class TestCaseContext:
+class TestCaseContext(Generic[TestCaseType]):
     """
     This class serves as a container for the user's test case +
     utilities around that test case.
     """
 
     # The user's test case
-    test_case: BaseTestCase
+    test_case: TestCaseType
 
     # Defined if the user has configured their test case to repeat
     repetition_idx: Optional[int]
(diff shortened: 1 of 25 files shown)


autoblocks/api/client.py
```diff
@@ -1 +1,5 @@
-from autoblocks._impl.api.client import AutoblocksAPIClient  # noqa: F401
+from autoblocks._impl.api.client import AutoblocksAPIClient
+
+__all__ = [
```
nicolewhite, line 3: fixes https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

autoblocks/_impl/testing/models.py
```diff
@@ -71,7 +79,7 @@ def hash(self) -> str:
         return self._cached_hash
 
 
-class BaseTestEvaluator(abc.ABC):
+class BaseTestEvaluator(abc.ABC, Generic[TestCaseType, OutputType]):
```
nicolewhite, line 82: evaluators need generics for test case and output, so this now matches how it's set up in TS

tests/autoblocks/test_prompt_manager.py
```diff
@@ -36,11 +41,11 @@ def my_template(
 
 class MyExecutionContext(
     PromptExecutionContext[
-        None,
```
nicolewhite, line 39: couldn't figure out how to do optional generics so just switched back to an empty pydantic class

adamnolte (approved):
Can we add a check here to try to catch these? Or i guess we could do that in the examples repo too?

tests/autoblocks/test_run_test_suite.py
```diff
@@ -317,7 +317,7 @@ def test_error_in_evaluator(httpx_mock):
     def test_fn(test_case: MyTestCase):
         return test_case.input + "!"
 
-    class MySyncEvaluator(BaseTestEvaluator):
+    class MySyncEvaluator(BaseTestEvaluator[MyTestCase, str]):
```
nicolewhite, line 320: they'll need to specify generics for their evaluators like this in order to get their mypy checks to pass

nicolewhite:
@adamnolte yeah looking at that now. this went undetected because we're not running on the `tests/` directory in this repo but i need to figure out how to do `--strict` but with some flags relaxed, like `--no-untyped-def`

nicolewhite:
@adamnolte added here: https://github.com/autoblocksai/python-sdk/pull/161/commits/5beffe88632e049e2754fa248663e2d665d58337

dmorton2297 (approved):

PR #176, opened 2024-04-28
Fix undeployed e2e test

nicolewhite:


tests/e2e/test_e2e.py
```diff
@@ -240,19 +239,44 @@ def test_prompt_manager_no_model_params():
         )
 
 
-def test_prompt_manager_no_model_params_undeployed():
+def test_prompt_manager_undeployed_latest_revision():
     mgr = UsedByCiDontDeleteNoParamsUndeployedPromptManager(
```
nicolewhite, line 243: still need to update the autogeneration scripts to not create a separate "Undeployed" class but will do that separately

tests/e2e/test_e2e.py
```diff
@@ -240,19 +239,44 @@ def test_prompt_manager_no_model_params():
...
-        UsedByCiDontDeleteNoParamsUndeployedMinorVersion.DANGEROUSLY_USE_UNDEPLOYED,
+        # Need to use a user-scoped API key to access undeployed prompts
+        api_key=os.environ["AUTOBLOCKS_API_KEY_USER"],
+        # Request the latest revision
+        minor_version="latest",
```
nicolewhite, line 247: the minor_version can actually be any string and we'll always grab the latest undeployed revision, but by setting it to "latest" it'll auto-update on an interval which might be nice for local dev if you're starting a flask app or something long-running

dmorton2297 (approved):

tests/e2e/test_e2e.py
```diff
@@ -240,19 +239,44 @@ def test_prompt_manager_no_model_params():
...
+
+def test_prompt_manager_undeployed_specific_revision():
+    """
+    This test uses a revision created in our CI org:
+
+    https://app.autoblocks.ai/prompts/used-by-ci-dont-delete-no-params/revisions/clvifur81000336hksycpw4av/edit
+    """
+    mgr = UsedByCiDontDeleteNoParamsUndeployedPromptManager(
+        # Need to use a user-scoped API key to access undeployed prompts
+        api_key=os.environ["AUTOBLOCKS_API_KEY_USER"],
```
nicolewhite, line 263: end-users won't need to have a separate env var with the _USER suffix, i just have it set up this way for our e2e tests so we can test both an org-scoped and user-scoped API key

end-users will just have an `AUTOBLOCKS_API_KEY` env var that is set to their org-scoped API key in prod and to user-scoped API keys in local dev

PR #183, opened 2024-04-29
EPD-846: Config SDK

adamnolte:


autoblocks/_impl/configs/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+@dataclass
+class DangerouslyUseUndeployedWithRevision:
+    revisionId: str
+
+
+@dataclass
+class DangerouslyUseUndeployedRemoteConfig:
+    id: str
+    dangerouslyUseUndeployed: Union[LatestDangerouslyUseUndeployed, DangerouslyUseUndeployedWithRevision]
```
nicolewhite, line 40:
```suggestion
    dangerously_use_undeployed: Union[LatestDangerouslyUseUndeployed, DangerouslyUseUndeployedWithRevision]
```

autoblocks/_impl/configs/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@dataclass
+class LatestDangerouslyUseUndeployed:
+    latest: Literal[True]
+
+
+@dataclass
+class DangerouslyUseUndeployedWithRevision:
+    revisionId: str
```
nicolewhite, line 34:
```suggestion
    revision_id: str
```

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+
+async def get_remote_config(config: RemoteConfig, timeout: timedelta, api_key: str) -> RemoteConfigResponse:
+    resp = await global_state.http_client().get(
+        make_request_url(config=config),
+        timeout=timeout.total_seconds(),
+        headers={"Authorization": f"Bearer {api_key}"},
+    )
+    resp.raise_for_status()
+    return RemoteConfigResponse(
+        id=resp.json().get("id"),
```
nicolewhite, line 97: are you able to call resp.json() multiple times?

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+from autoblocks._impl.configs.models import RemoteConfig
+from autoblocks._impl.configs.models import RemoteConfigResponse
+from autoblocks._impl.configs.models import RemoteConfigWithVersion
+from autoblocks._impl.util import AnyTask
+from autoblocks._impl.util import AutoblocksEnvVar
+from autoblocks._impl.util import get_running_loop
+
+log = logging.getLogger(__name__)
+
+AutoblocksConfigValue = TypeVar("AutoblocksConfigValue")
```
nicolewhite, line 28: maybe `AutoblocksConfigValueType`

autoblocks/_impl/configs/config.py
adamnolte: hmm ya i was able too (the test passed), but i'll just pull it out into a var

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+
+        If the parser is specified, the value will be parsed using the parser.
+        """
+        remote_config = await get_remote_config(config, timeout=timeout, api_key=api_key)
+        try:
+            parsed = parser(remote_config.value)
+            if parsed is not None:
+                self._value = parsed
+        except Exception as e:
+            raise Exception(f"Failed to parse config '{config.id}': {e}")
```
nicolewhite, line 156: i would log an error but re-raise the original instead of turning it into a new exception:

```python
except Exception as e:
    log.error(f"Failed to parse config '{config.id}': {e}", exc_info=True)
    raise
```

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+                refresh_interval=refresh_interval,
+                refresh_timeout=refresh_timeout,
+                activate_timeout=activate_timeout,
+                parser=parser,
+            )
+        except Exception as err:
+            log.error(f"Failed to activate remote config '{config.id}': {err}")
+
+    @property
+    def value(self) -> AutoblocksConfigValue:
```
nicolewhite: do we need to put access behind a context manager so we can reliably track what version was used during a request?

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+        return {}
+    return json.loads(config_revisions_raw)  # type: ignore
+
+
+def make_request_url(config: RemoteConfig) -> str:
+    config_id = config.id
+    base = f"{API_ENDPOINT}/configs/{config_id}"
+    revision_id = config_revisions_map().get(config_id)
+    if revision_id and is_testing_context():
+        return f"{base}/revisions/{revision_id}"
```
nicolewhite, line 62: super nit:

```python
if is_testing_context() and revision_id := config_revisions_map().get(config_id):
    return f"{base}/revisions/{revision_id}"
```

just so you don't even hit the `config_revisions_map().get(config_id)` code unless you're in a testing context

autoblocks/_impl/configs/config.py
```diff
@@ -0,0 +1,246 @@
...
+        make_request_url(config=config),
+        timeout=timeout.total_seconds(),
+        headers={"Authorization": f"Bearer {api_key}"},
+    )
+    resp.raise_for_status()
+    resp_json = resp.json()
+    return RemoteConfigResponse(
+        id=resp_json.get("id"),
+        version=resp_json.get("version"),
+        value=resp_json.get("value"),
```
nicolewhite, line 97: should we access directly here instead of using get? might be preferable to have a keyerror here rather than let it continue in an invalid state?

adamnolte: ya makes sense

tests/autoblocks/test_autoblocks_configs.py
```diff
@@ -0,0 +1,249 @@
...
+                random_val="val-from-remote",
+            ),
+        ),
+    )
+
+    config = MyConfig(
+        value=MyConfigValue(my_val="initial-val"),
+    )
+    config.activate_from_remote(
+        config=LatestRemoteConfig(id="my-config-id", latest=True), parser=MyConfigValue.model_validate
```
nicolewhite, line 49: should we default latest=True on `LatestRemoteConfig` so you don't have to repeat it? so it'd just be:

```python
config.activate_from_remote(
  config=LatestRemoteConfig(id="my-config-id"),
  ...,
)
```

since it's already in the name that it's latest

tests/autoblocks/test_autoblocks_configs.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+
+    config = MyConfig(
+        value=MyConfigValue(my_val="initial-val"),
+    )
+    config.activate_from_remote(
+        config=DangerouslyUseUndeployedRemoteConfig(
+            id="my-config-id",
+            dangerously_use_undeployed=DangerouslyUseUndeployedWithRevision(revision_id="my-revision-id"),
+        ),
```
nicolewhite, line 175: seeing this written out it's pretty verbose, and will also require you to make quite a few temp code changes locally to use a local revision

adamnolte: ya i was thinking that too for python we could do it all top level so like

DangerouslyUseUndeployedRemoteConfigWithRevision ?

nicolewhite (dismissed):

nicolewhite (dismissed):

tests/autoblocks/test_autoblocks_configs.py
adamnolte: update this, I think it does feel nicer, lmk what you think

PR #184, opened 2024-04-29
EPD-864 Collect revision usage during test case run

nicolewhite:


autoblocks/_impl/context_vars.py
```diff
@@ -1,14 +1,67 @@
...
+
+def register_test_case_revision_usage(
+    entity_id: str,
+    entity_type: RevisionType,
+    revision_id: str,
+) -> None:
+    ctx = test_case_run_context_var.get()
+    if ctx is None:
+        # Shouldn't happen, but just return
+        return
```
dmorton2297, line 53: Maybe return an empty list so you don't have to do check for None

```
testCaseRevisionUsage=[usage.serialize() for usage in revision_usage] if revision_usage else None,
```

in run.py

tests/autoblocks/test_run_test_suite.py
```diff
@@ -1743,3 +1776,173 @@ def test_run_stops_if_start_fails(httpx_mock):
...
+            testCaseOutput="z!",
+            testCaseDurationMs=ANY_NUMBER,
+            testCaseRevisionUsage=[
+                dict(
+                    entityExternalId="prompt-a",
+                    entityType="prompt",
+                    revisionId="prompt-a-mock-revision-id",
+                    usedAt=mock.ANY,
+                ),
+            ],
```
dmorton2297, line 1927: Nice!

dmorton2297 (approved):

autoblocks/_impl/context_vars.py
nicolewhite: i think i want it to be None rather than an empty list when we send it to our API

nicolewhite: also this is register_test_case_revision_usage not get_test_case_revision_usage

PR #186, opened 2024-05-02
Update prompts generate CLI to handle revisions

nicolewhite:
* dropped minor version enum (since we're now just requesting the prompts in the yaml config instead of all of them), i think was kind of annoying to import & use anyway b/c you had to run `prompts generate` just to bump the minor version
* updated the schema of the yaml file to match how configs are fetched (you can specify a major version or dangerously_use_undeployed_revision)
* switched to snapshot testing for testing the autogenerated code

### DIFF
--- a/.autoblocks.yml
+++ b/.autoblocks.yml
@@ -4,9 +4,13 @@ autogenerate:
 
     prompts:
       - id: used-by-ci-dont-delete
-        major_versions:
-          - 2
+        major_version: 2
 
       - id: used-by-ci-dont-delete-no-params
-        major_versions:
-          - 1
+        major_version: 1
+
+      - id: text-summarization
+        dangerously_use_undeployed_revision: latest
+
+      - id: question-answerer
+        dangerously_use_undeployed_revision: clvodtv700003a2z02fumceby
--- a/autoblocks/_impl/prompts/autogenerate.py
+++ b/autoblocks/_impl/prompts/autogenerate.py
@@ -1,4 +1,3 @@
-import logging
 import re
 from typing import Any
 from typing import Dict
@@ -13,8 +12,6 @@ from autoblocks._impl.config.constants import REVISION_UNDEPLOYED
 from autoblocks._impl.prompts.models import AutogeneratePromptsConfig
 from autoblocks._impl.prompts.models import FrozenModel
 from autoblocks._impl.util import AutoblocksEnvVar
-
-log = logging.getLogger(__name__)
 
 
 class TemplateParam(FrozenModel):
@@ -253,12 +250,10 @@ def make_prompts_from_config(
             major = REVISION_UNDEPLOYED
             minor = prompt.dangerously_use_undeployed_revision
         else:
-            log.error(
+            raise ValueError(
                 f"Error in {prompt.id} config: "
-                f"Either `major_version` or `dangerously_use_undeployed_revision` must be specified."
-                f"Skipping prompt."
+                f"Either `major_version` or `dangerously_use_undeployed_revision` must be specified"
             )
-            continue
 
         resp = httpx.get(
             url=f"{API_ENDPOINT}/prompts/{prompt.id}/major/{major}/minor/{minor}",
--- a/autoblocks/_impl/prompts/autogenerate.py
+++ b/autoblocks/_impl/prompts/autogenerate.py
@@ -1,3 +1,4 @@
+import logging
 import re
 from typing import Any
 from typing import Dict
@@ -12,6 +13,8 @@ from autoblocks._impl.config.constants import REVISION_UNDEPLOYED
 from autoblocks._impl.prompts.models import AutogeneratePromptsConfig
 from autoblocks._impl.prompts.models import FrozenModel
 from autoblocks._impl.util import AutoblocksEnvVar
+
+log = logging.getLogger(__name__)
 
 
 class TemplateParam(FrozenModel):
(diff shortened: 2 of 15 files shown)


.autoblocks.yml
```diff
@@ -4,9 +4,13 @@ autogenerate:
...
       - id: used-by-ci-dont-delete-no-params
-        major_versions:
-          - 1
+        major_version: 1
+
+      - id: text-summarization
+        dangerously_use_undeployed_revision: latest
+
+      - id: question-answerer
+        dangerously_use_undeployed_revision: clvodtv700003a2z02fumceby
```
nicolewhite, line 16: what's still a little awkward is that if you want to both autogenerate + use a specific revision you have to specify it here (to generate the classes) and again when initializing the manager (via the minor_version arg), i think can do something nicer in a followup

.autoblocks.yml
```diff
@@ -4,9 +4,13 @@ autogenerate:
 
     prompts:
       - id: used-by-ci-dont-delete
-        major_versions:
-          - 2
+        major_version: 2
 
       - id: used-by-ci-dont-delete-no-params
-        major_versions:
```
nicolewhite, line 11: switching from a list to a single minor version. i think supporting a list here was overly complicated and an unlikely edge case. can reconsider if someone actually wants to use multiple major versions of the same prompt simultaneously

autoblocks/_impl/prompts/autogenerate.py
```diff
@@ -222,86 +208,64 @@ def generate_execution_context_class_code(prompt: Prompt) -> str:
...
-    for (prompt_id, major_version), (row, minor_versions) in by_major_version.items():
+    for prompt in config.prompts:
+        if prompt.major_version is not None:
+            major = prompt.major_version
+            minor = REVISION_LATEST
+        elif prompt.dangerously_use_undeployed_revision is not None:
+            major = REVISION_UNDEPLOYED
+            minor = prompt.dangerously_use_undeployed_revision
+        else:
+            raise ValueError("Either major_version or dangerously_use_undeployed_revision must be specified")
```
adamnolte, line 253: should we just skip in this case and log rather than raising an exception?

adamnolte: maybe log which prompt id too

nicolewhite: yeah i'll add the prompt ID but i feel it should throw right? rather than finishing and letting you discover on your own that one of your prompts is missing

adamnolte: ya I think throwing makes sense thinking more about it.

autoblocks/_impl/prompts/autogenerate.py
```diff
@@ -222,86 +208,67 @@ def generate_execution_context_class_code(prompt: Prompt) -> str:
...
-        if key not in by_major_version:
-            by_major_version[key] = (row, [])
-        by_major_version[key][1].append(minor_version)
+    for prompt in config.prompts:
+        if prompt.major_version is not None:
+            major = prompt.major_version
+            minor = REVISION_LATEST
+        elif prompt.dangerously_use_undeployed_revision is not None:
+            major = REVISION_UNDEPLOYED
+            minor = prompt.dangerously_use_undeployed_revision
```
dmorton2297, line 251: 👍 this clears up my comment over on the docs pr https://github.com/autoblocksai/autoblocks-docs/pull/235#pullrequestreview-2036497132

autoblocks/_impl/prompts/autogenerate.py
```diff
@@ -66,19 +57,14 @@ def execution_context_class_name(self) -> str:
...
 
 HEADER = """############################################################################
 # This file was generated automatically by Autoblocks. Do not edit directly.
 ############################################################################
 
-from enum import Enum
 from typing import Union  # noqa: F401
 
-import pydantic
+import pydantic  # noqa: F401
```
dmorton2297, line 67: Why do we need these noqa tags here?

tests/autoblocks/test_prompt_manager.py
```diff
@@ -49,16 +48,8 @@ class MyExecutionContext(
...
-    v0 = "0"
-    LATEST = "latest"
-
-
 class MyPromptManager(
-    AutoblocksPromptManager[
-        MyExecutionContext,
-        MyMinorVersion,
-    ],
+    AutoblocksPromptManager[MyExecutionContext,],
```
dmorton2297, line 52:
```suggestion
    AutoblocksPromptManager[MyExecutionContext],
```

tests/autoblocks/test_run_test_suite.py
```diff
@@ -1800,25 +1799,15 @@ class MyExecutionContext(
...
-    class MyMinorVersion(Enum):
-        v0 = "0"
-        LATEST = "latest"
-
     class PromptManagerA(
-        AutoblocksPromptManager[
-            MyExecutionContext,
-            MyMinorVersion,
-        ],
+        AutoblocksPromptManager[MyExecutionContext,],
```
dmorton2297, line 1803: Does the formatter add these commas?

autoblocks/_impl/prompts/autogenerate.py
nicolewhite: b/c pydantic might not be used if you don't use params, so if we don't noqa it our autogenerated file will fail flake8 unused imports rule. could probably add the imports retroactively instead so we don't need to add noqa to things

tests/autoblocks/test_run_test_suite.py
nicolewhite: no that's black reformatting (it was originally like this):

```python

AutoblocksPromptManager[
  MyExecutionContext,
],

adamnolte (approved):

PR #187, opened 2024-05-05
update python sdk for configs updates

adamnolte:


autoblocks/_impl/configs/models.py
```diff
@@ -1,19 +1,38 @@
...
+    properties: List[Property]
 
 
 @dataclass
 class RemoteConfig:
     id: str
     # only one of these should be specified
-    version: Optional[str] = None  # can be a version like "1" or "latest"
+    major_version: Optional[int] = None  # should be a sepcific major version
+    minor_version: Optional[int] = None  # omitted for latest minor version
```
nicolewhite, line 37: can we match prompts here? where minor_version type is `str` and is required

nicolewhite (dismissed):

autoblocks/_impl/configs/models.py
nicolewhite: i guess that will only work if you make major/minor its own object though. but i still think it would be nicer to make ppl be explicit about the minor version. for prompts you have to do:

```python
mgr = Manager(minor_version="0")

# or

mgr = Manager(minor_version="latest")
```

PR #191, opened 2024-05-06
EPD-885 Add before-evaluators hook

nicolewhite:


autoblocks/_impl/testing/run.py
```diff
@@ -173,6 +186,26 @@ async def run_test_case_unsafe(
                 test_case_ctx.test_case,
             )
 
+        # Run the before-evaluators hook if provided.
+        # Note we run this within the test case semaphore so that
+        # max_test_case_concurrency applies to both fn + before_evaluators_hook.
+        hook_results = None
+        if before_evaluators_hook:
+            if inspect.iscoroutinefunction(before_evaluators_hook):
```
dmorton2297, line 194: 🎉

dmorton2297: Just to remind me - will this block if the hook function fails?

dmorton2297 (approved):

autoblocks/_impl/testing/run.py
nicolewhite: yeah

PR #214, opened 2024-06-24
Make OOB evals as ABC

adamnolte:
@nicolewhite I think this does feel nicer, wdyt?

### DIFF
--- a/autoblocks/_impl/testing/evaluators/battle.py
+++ b/autoblocks/_impl/testing/evaluators/battle.py
@@ -1,15 +1,15 @@
+import abc
 import json
 from dataclasses import dataclass
 from textwrap import dedent
-from typing import Callable
-from typing import Generic
+from typing import Any
 from typing import Optional
 
 from autoblocks._impl import global_state
 from autoblocks._impl.config.constants import API_ENDPOINT
-from autoblocks._impl.context_vars import test_case_run_context_var
 from autoblocks._impl.testing.evaluators.util import get_autoblocks_api_key
 from autoblocks._impl.testing.evaluators.util import get_openai_client
+from autoblocks._impl.testing.evaluators.util import get_test_id
 from autoblocks._impl.testing.models import BaseTestEvaluator
 from autoblocks._impl.testing.models import Evaluation
 from autoblocks._impl.testing.models import OutputType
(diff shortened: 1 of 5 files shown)


autoblocks/_impl/testing/evaluators/battle.py
```diff
@@ -23,37 +23,130 @@ class BattleResponse:
...
+    @abc.abstractmethod
+    def criteria(self) -> str:
+        pass
 
-    async def get_baseline(self, test_id: str, test_case: TestCaseType) -> Optional[str]:
-        # if the user passed in a baseline_mapper, we use that
-        if self.baseline_mapper:
-            return self.baseline_mapper(test_case)
+    @abc.abstractmethod
+    def baseline_mapper(self, test_case: Any) -> str:
```
nicolewhite, line 110: you can keep the generics so that this can be typed `test_case: TestCaseType`

we only had to drop it for `evaluate_test_case` b/c we needed to support kwargs for hook results

autoblocks/_impl/testing/evaluators/has_all_substrings.py
```diff
@@ -9,25 +9,26 @@
...
+class HasAllSubstrings(BaseTestEvaluator, abc.ABC):
     """
     The HasAllSubstrings evaluator checks if the output contains all the expected substrings.
     Scores 1 if all substrings are present, 0 otherwise.
     The comparison is case-sensitive.
     """
 
-    id = "has-all-substrings"
+    @abc.abstractmethod
+    def test_case_mapper(self, test_case: Any) -> List[str]:
```
nicolewhite, line 20: i think we can keep the generics here as well

adamnolte: I was getting a mypy error with the generics when I tried to add it in the test

error: Argument 1 of "output_mapper" is incompatible with supertype "ManualBattle"; supertype defines the argument type as "OutputType"  [override]
 note: This violates the Liskov substitution principle
note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

autoblocks/_impl/testing/evaluators/battle.py
```diff
@@ -23,37 +23,130 @@ class BattleResponse:
...
+
+    async def evaluate_test_case(self, test_case: TestCaseType, output: OutputType) -> Evaluation:
+        mapped_output = self.output_mapper(output)
+        baseline = self.baseline_mapper(test_case)
+
+        return await battle(baseline=baseline, challenger=mapped_output, criteria=self.criteria, evaluator_id=self.id)
 
-        # otherwise, we fetch the baseline from the API
+
+class AutomaticBattle(BaseTestEvaluator, abc.ABC):
```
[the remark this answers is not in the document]
adamnolte: automatic, we save the baseline for you. manual you provide the baseline yourself from your test cases.
open to naming suggestions.

the other option would be to keep it the same and just check if they implemented baseline_mapper like we were doing before.

autoblocks/_impl/testing/evaluators/battle.py
```diff
@@ -23,37 +23,130 @@ class BattleResponse:
...
+        pass
+
+    @abc.abstractmethod
+    def output_mapper(self, output: Any) -> str:
+        """
+        Map your output to a string for comparison to the baseline.
+        """
+        pass
+
+    async def evaluate_test_case(self, test_case: TestCaseType, output: OutputType) -> Evaluation:
```
nicolewhite, line 123: should the signature here be ` test_case: Any, output: Any` to match the above since there are no generics?

autoblocks/_impl/testing/evaluators/has_all_substrings.py
nicolewhite: hmm this isn't extending ManualBattle so maybe that error was coming from something else?

adamnolte: It was coming from the test

```
    class Battle(ManualBattle):
        id = "battle"
        criteria = "Choose the best greeting."

        def output_mapper(self, output: str) -> str:
            return output

        def baseline_mapper(self, test_case: MyTestCase) -> str:
            return "goodbye world"
  ```

File: autoblocks/_impl/testing/evaluators/battle.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (nicolewhite) [reply autoblocks/_impl/testing/evaluators/battle.py:130]: ok the latter sounds nicer i think

File: autoblocks/_impl/testing/evaluators/has_all_substrings.py (hunk 2 above)
Comment (nicolewhite) [reply autoblocks/_impl/testing/evaluators/has_all_substrings.py:20]: yeah i am commenting on the HasAllSubstrings class here, but if this is only used in tests it's fine

File: autoblocks/_impl/testing/evaluators/battle.py [hunk 5]
```diff
@@ -143,25 +175,13 @@ async def evaluate_test_case(self, test_case: TestCaseType, output: OutputType)
             await self.save_baseline(test_id=test_id, baseline=mapped_output, test_case_hash=test_case.hash())
             return Evaluation(
```
nicolewhite, line 176: we could also return `None` here, which would follow the comment more closely where it says "We save the current challenger as the baseline and skip evaluating"

nicolewhite (dismissed):

autoblocks/_impl/testing/evaluators/battle.py
adamnolte: oh ya, i actually had that orginally but switched to this since we didn't have that in js sdk yet. let me switch it back then i'll update js sdk to support none

nicolewhite (approved):

PR #216, opened 2024-06-25
EPD-848 grid search params

nicolewhite:


autoblocks/_impl/context_vars.py
```diff
@@ -53,12 +56,19 @@ class EvaluatorRunContext:
 )
 
 
+def get_current_run_context() -> Optional[Union[EvaluatorRunContext, TestCaseRunContext]]:
+    # Check the narrower evaluator_run_context first so that we can detect if a prompt
+    # is being used in the context of an evaluator. Otherwise, fall back to the wider
+    # test_case_run_context.
+    return evaluator_run_context_var.get() or test_case_run_context_var.get()
```
adamnolte, line 63: so i actually changed the behavior of the test case run context to be around evaluators as well so i could grab the test id.

I guess the bug that would create is there is no separation of eval revisions and app revisions?
I can switch that back and just add test id to the evaluator run context - https://github.com/autoblocksai/python-sdk/pull/209/files#diff-7ee7680f9f0b6b9dcc3bc6d6479bddde5790866ea82b180e4ad772be232f7babR299

nicolewhite: this isn't changing anything just FYI, just pulling this into a function. or are you just saying you want to revert your change ?

adamnolte: i'm just saying i should revert my change

nicolewhite: ok i'm not sure what the problem is exactly but sounds good

adamnolte: I guess it actually won't be a bug because we check evaluator_run_context first

nicolewhite: yeah your change doesn't affect how we store and then retrieve revision usage

adamnolte (approved):

adamnolte (approved):

adamnolte (approved):

PR #282, opened 2024-10-17
python-sdk add methods to get run ids, result ids and test case result

nenadnovovic-ab:
## Summary by CodeRabbit

- **New Features**
	- Enhanced API client with new methods for retrieving local and CI test runs and their results.
	- Introduction of new data models for test runs and evaluations.

- **Bug Fixes**
	- Improved validation and retrieval processes for test case results.

- **Tests**
	- Added new test functions to validate the functionality of the API client methods.


autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,33 @@ class Dataset:
...
+@dataclass
+class Evaluation:
+    evaluatorId: str
+    score: float
+    passed: bool
+    metadata: Dict[str, Any]
+    threshold: Optional[Dict[str, Optional[float]]] = None
+
+@dataclass
+class AutoblocksTestCaseResultWithEvaluations:
```
adamnolte, line 224: let's drop the `WithEvaluations`

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,33 @@ class Dataset:
...
+    id: str
+    runId: str
+    hash: str
+    datasetItemId: Optional[str]
```
adamnolte, line 228:
```suggestion
    dataset_item_id: Optional[str]
```

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,33 @@ class Dataset:
...
+    durationMs: Optional[int]
```
adamnolte, line 229: duration_ms

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +233,44 @@ def get_dataset(
...
+        ]
+
+    def get_ci_test_run_results(self, run_id: str) -> List[AutoblocksTestCaseResult]:
+        resp = self._client.get(f"/testing/ci/runs/{encode_uri_component(run_id)}/results")
+        data = resp.json()  # Parse the response JSON
+        return [AutoblocksTestCaseResult(
+            id=result["id"]
+        ) for result in data["results"]]
+
+    def get_local_test_case_result(self, test_case_result_id: str) -> AutoblocksTestCaseResultWithEvaluations:
```
adamnolte, line 266: let's make sure we are consisenst between the js and python sdk so here we have `get_local_test_case_result` but in the js sdk we call it test result.

I think on both we should do test case result.

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +233,44 @@ def get_dataset(
...
+        data = resp.json()
+        return [AutoblocksTestCaseResult(
+                id=result["id"]
+            )
+            for result in data["results"]
+        ]
+
+    def get_ci_test_run_results(self, run_id: str) -> List[AutoblocksTestCaseResult]:
+        resp = self._client.get(f"/testing/ci/runs/{encode_uri_component(run_id)}/results")
+        data = resp.json()  # Parse the response JSON
```
adamnolte, line 261: i think you are missing a raise for status

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +233,44 @@ def get_dataset(
...
+
+    def get_ci_test_runs(self, test_external_id: str) -> List[AutoblocksTestRun]:
+        req = self._client.get(f"/testing/ci/tests/{encode_uri_component(test_external_id)}/runs")
+        req.raise_for_status()
+        resp = req.json()
+        return [AutoblocksTestRun(id=run["id"]) for run in resp["runs"]]
+
+    def get_local_test_run_results(self, run_id: str) -> List[AutoblocksTestCaseResult]:
+        resp = self._client.get(f"/testing/local/runs/{encode_uri_component(run_id)}/results")
+        resp.raise_for_status()
```
adamnolte, line 251: let's be consisistent with the naming, should be req = self._client

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +234,89 @@ def get_dataset(
...
+        # Convert evaluations to Evaluation objects
+        evaluations = [
+            Evaluation(
+                evaluatorId=eval.get("evaluatorId"),
+                score=eval.get("score"),
+                passed=eval.get("passed"),
+                metadata=eval.get("metadata", {}),
+            )
+            for eval in result.get("evaluations", [])
+        ]
```
adamnolte, line 289: we need threshold here right?

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +234,89 @@ def get_dataset(
...
```
adamnolte, line 320: same with here for threshold?

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +234,89 @@ def get_dataset(
...
+
+        return AutoblocksTestCaseResult(**{**result, "events": events, "evaluations": evaluations})
```
adamnolte, line 322: I think duration, run id, and dataset item id are going to fail here because the api returns like "durationMs", so we need to map them explicitly. Let's just do that for all the fields

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,35 @@ class Dataset:
...
+
+
+@dataclass
+class AutoblocksTestCaseResultId:
+    id: str
+
+
+@dataclass
+class Evaluation:
+    evaluatorId: str
```
adamnolte, line 218:
```suggestion
    evaluator_id: str
```

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,35 @@ class Dataset:
...
+    score: float
+    passed: bool
+    metadata: Dict[str, Any]
+    threshold: Optional[Dict[str, Optional[float]]] = None
+
+
+@dataclass
+class AutoblocksTestCaseResult:
+    id: str
+    runId: str
```
adamnolte, line 228:
```suggestion
    run_id: str
```

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,35 @@ class Dataset:
...
+    id: str
+
+
+@dataclass
+class Evaluation:
+    evaluatorId: str
+    score: float
+    passed: bool
+    metadata: Dict[str, Any]
+    threshold: Optional[Dict[str, Optional[float]]] = None
```
adamnolte, line 222: are we returning assertions from the public api? If not we should and also set that in both js sdk and python sdk

autoblocks/_impl/api/client.py
nenadnovovic-ab: Ok, I also updated the tests so the mock json returns the camel case like runId, and the actual result expects the run_id.

autoblocks/_impl/api/client.py
```diff
@@ -230,3 +235,129 @@ def get_dataset(
...
+    def get_local_test_result(self, test_case_result_id: str) -> AutoblocksTestCaseResult:
+        req = self._client.get(f"/testing/local/results/{encode_uri_component(test_case_result_id)}")
+        req.raise_for_status()
+        resp = req.json()
+        result = resp["testCaseResult"]
+
+        # Convert events to Event objects
+        events = [
+            Event(
+                id=event.get("id"),
```
adamnolte, line 272: we should update the usages of get to only when we expect something to be able to be set for None, that way we through an error when something unexpected happens.

For example id should always be there, but properties we can keep as get with a default. Same with the other usages of `.get`

nenadnovovic-ab: @adamnolte so, do you mean for example, as we always expect id, i do
 id=event["id"] instead, and for all the other non-optional fields

autoblocks/_impl/api/client.py
```diff
@@ -281,17 +281,17 @@ def get_local_test_result(self, test_case_result_id: str) -> AutoblocksTestCaseR
...
         evaluations = [
             EvaluationWithEvaluatorId(
-                evaluator_id=eval.get("evaluatorId"),
-                score=eval.get("score"),
-                passed=eval.get("passed"),
-                metadata=eval.get("metadata", {}),
-                threshold=eval.get("threshold"),
+                evaluator_id=eval["evaluatorId"],
+                score=eval["score"],
+                passed=eval["passed"],
```
adamnolte, line 286: this should actually stay as .get since its optional - https://github.com/autoblocksai/autoblocks/blob/main/apps/public-api/src/testing/types.ts#L43

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,44 @@ class Dataset:
...
+    required: bool
+    criterion: str
+    metadata: Optional[Dict[str, Any]] = None
+
+
+@dataclass
+class EvaluationWithEvaluatorId:
+    evaluator_id: str
+    score: float
+    passed: bool
```
adamnolte, line 228:
```suggestion
    passed: Optional[bool]
```

autoblocks/_impl/api/models.py
```diff
@@ -201,3 +201,44 @@ class Dataset:
...
+    metadata: Dict[str, Any]
```
adamnolte, line 229:
```suggestion
    metadata: Optional[Dict[str, Any]]
```

adamnolte (approved):

PR #335, opened 2025-05-02
Add DatasetsV2Client

josorio-autoblocks:
## Greptile Summary

This PR introduces a comprehensive DatasetsV2Client implementation with models, utilities, and testing for the Autoblocks Python SDK. Here are the key points to review:

- The `DatasetsClient` in `autoblocks/_impl/datasets/client.py` needs proper resource cleanup in the `__init__` method for the `httpx.Client` to prevent connection leaks
- The schema validation in `autoblocks/_impl/datasets/models/schema.py` lacks proper validation for empty role lists in `ConversationProperty` 
- The `build_path` function in `autoblocks/_impl/datasets/utils/helpers.py` requires URL encoding for query parameters and path segments to handle special characters
- The error handling in `autoblocks/_impl/datasets/exceptions.py` uses bare `except Exception` blocks that should catch specific exceptions
- The test coverage in `tests/e2e/test_datasets.py` needs expansion to include error cases and edge conditions

The implementation is well-structured but needs these improvements for production readiness. The code follows Python best practices with proper type hints and validation logic.


josorio-autoblocks:
@greptile

autoblocks/_impl/api/app_client.py
```diff
@@ -0,0 +1,60 @@
...
+
+class AutoblocksAppClient:
+    """
+    API client for Autoblocks app-specific operations.
+
+    This client is used for operations that require an app_slug, such as
+    working with datasets.
+    """
+
+    def __init__(self, app_slug: str, api_key: Optional[str] = None, timeout_ms: int = 60000) -> None:
```
adamnolte, line 20: timeout should use the timedelta class - https://github.com/autoblocksai/python-sdk/blob/main/autoblocks/_impl/api/client.py#L81

autoblocks/_impl/datasets/exceptions.py
```diff
@@ -0,0 +1,272 @@
+"""Custom exceptions for Datasets API."""
```
adamnolte, line 1: are these specific to datasets? They seem general?

josorio-autoblocks: I created them for this specific use case, if we create more api clients in the future, we should move and reuse

autoblocks/_impl/api/app_client.py
josorio-autoblocks: done, ty

tests/e2e/test_prompts_v2.py
```diff
@@ -30,27 +28,26 @@ def test_app_sdk_test_prompt_basic_v1():
...
     with mgr.exec() as ctx:
         # Assert parameters
         assert ctx.params.model == "gpt-4o"
-        assert ctx.params.max_tokens == 256
+        assert ctx.params.max_tokens == 256  # type: ignore[union-attr]
 
         # Test template rendering
         assert (
-            ctx.render_template.template_c(
+            ctx.render_template.template_c(  # type: ignore[union-attr]
```
adamnolte, line 50: what are these type ignores for?

adamnolte (dismissed):

tests/e2e/test_prompts_v2.py
josorio-autoblocks: in this case, we return a union type, our current setup complains because in one type template_c is not there

autoblocks/_impl/api/app_client.py
```diff
@@ -0,0 +1,60 @@
...
+
+        self.app_slug = app_slug
+        self.timeout = timeout
+
+        # Initialize datasets client lazily
+        self._datasets = DatasetsClient(
+            {
+                "api_key": self.api_key,
+                "app_slug": self.app_slug,
+                "timeout_ms": int(timeout.total_seconds() * 1000),
```
adamnolte, line 48: this seems weird that we convert from seconds to milliseconds then in the datasets client we convert back to seconds for httpx. can we just pass around the time delta then use seconds with httpx rather than doing these conversions?

adamnolte (approved):

